PR #2, opened 2021-08-22
switch hello-world app to typescript

vkarpov15:
## What was changed

As per discussion last week, started refactoring samples to use TypeScript first.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (commented):
I'd rather we move the hello world sample completely out of the `sdk-node` repo than have this duplication.
In order to achieve that we'd need:
- CI with eslint, prettier and integration tests
- Triggers for CI cross repos
- Make the package initializer use samples from this repo

I've also mentioned in a comment using a single tsconfig file which will allow us to flatten the directory structure.

hello-world/src/worker/index.ts
```diff
@@ -1,7 +1,5 @@
-'use strict';
-
-// @@@SNIPSTART nodejs-js-hello-worker
-const { Worker } = require('@temporalio/worker');
+// @@@SNIPSTART nodejs-hello-worker
```
bergundy, line 1: Can you please copy all of the comments too? They're important for first time users and the snippets.
This goes for all of the files.

hello-world/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+  "compilerOptions": {
+    "emitDecoratorMetadata": false,
+    "experimentalDecorators": false,
+    "declaration": true,
+    "declarationMap": true,
+    "sourceMap": true,
+    "composite": true,
+    "incremental": true,
+    "rootDir": "./src",
+    "outDir": "./lib"
```
bergundy, line 13: You'll need to add the `@activities` alias here.
We don't really need our project to be composite.
You should probably exclude `node_modules`.

PR #4, opened 2021-08-27
Rewrite progress, http, expense examples in TS

vkarpov15:
## What was changed

progress and http examples use TypeScript now, similar to hello-world

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


expense/src/activities/payment.ts
```diff
@@ -0,0 +1,5 @@
+import axios from 'axios';
+
+export async function payment(id: string): Promise<void> {
```
bergundy, line 3: Is there a reason not to include both the activities in the same file?

expense/src/server/index.ts
```diff
@@ -1,21 +1,25 @@
-'use strict';
-
-const express = require('express');
+import { Connection, WorkflowClient } from '@temporalio/client';
+import express from 'express';
 
 const expenseState = Object.freeze({
```
bergundy, line 4: use typescript enum please

expense/src/server/index.ts
```diff
@@ -26,42 +30,54 @@ async function run() {
...
   app.post('/create', function(req, res) {
-    allExpenses[req.body.id] = expenseState.created;
+    allExpenses.set(req.body.id, expenseState.created);
 
     return res.json({ ok: 1 });
   });
 
   app.post('/action', function(req, res) {
-    const oldState = allExpenses[req.body.id];
+    const id = req.body.id;
```
bergundy, line 39: What do you think about this syntax?
```suggestion
    const { id } = req.body;
```

bergundy: Also, if you're not validating the request, at least add a comment that in "real code" this is required.

expense/src/server/index.ts
```diff
@@ -26,42 +30,54 @@ async function run() {
...
+      allExpenses.set(id, expenseState.completed);
       break;
     default:
       throw new Error(`Invalid action ${req.body.action}`);
     }
 
+    if (oldState === expenseState.created &&
+        (newState === expenseState.approved || newState === expenseState.rejected)) {
+      // TODO: need to be able to complete an activity here like in Go
+      // https://github.com/temporalio/samples-go/blob/0248987774da74dceb265bb3f3b06e9d00e50a32/expense/server/main.go#L166
```
bergundy, line 66: Just use `Connection.service` to do this: https://nodejs.temporal.io/api/classes/proto.temporal.api.workflowservice.v1.WorkflowService-1#respondactivitytaskcompleted

bergundy: We'll add a connection getter on activity context so you don't have to propagate the configuration to the activity.

expense/src/server/index.ts
```diff
@@ -26,42 +30,54 @@ async function run() {
...
 
   app.get('/status', function(req, res) {
     const id = req.query.id;
-    return res.json({ status: allExpenses[id] });
+    return res.json({ status: allExpenses.get(id) });
   });
 
   app.post('/registerCallback', function(req, res) {
     const id = req.body.id;
     const taskToken = req.body.taskToken;
```
bergundy, line 79: Same here:
```suggestion
    const { id, taskToken } = req.body;
```

expense/src/workflows/expense.ts
```diff
@@ -1,10 +1,8 @@
...
+import { createExpense } from '@activities/createExpense';
+import { payment } from '@activities/payment' ;
+import { sleep } from '@temporalio/workflow';
 
-const { createExpense } = require('@activities/createExpense');
-const { payment } = require('@activities/payment');
-const { sleep } = require('@temporalio/workflow');
-
-let complete = null;
+let complete: Function | null = null;
```
bergundy, line 5: Please use a `Trigger` instead.

progress/.gitignore
```diff
@@ -0,0 +1 @@
+lib
```
bergundy, line 1: Should we put all of the `lib` ignores in the root of the repo or move `node_modules` into each sample's gitignore?

progress/src/starter/index.ts
```diff
@@ -1,17 +1,16 @@
...
   const client = new WorkflowClient(connection.service);
 
-  const progress = client.stub('progress', { taskQueue: 'tutorial' });
+  const progress = client.stub<Progress>('progress', { taskQueue: 'tutorial20210827' });
 
   await progress.start();
 
   const val = await progress.query.getProgress();
-  console.log(val);
+  console.log(val); // 10
```
bergundy, line 13: Should be 10 if the timing is right

bergundy (commented):
I'm missing comments all of the samples, this is important because it really affects people's ability to learn from them.

progress/.gitignore
vkarpov15: I think we should have node_modules in each sample's `gitignore`. The idea is that each sample should be usable as a standalone, so we shouldn't rely on anything in the top-level dir.

expense/src/server/index.ts
vkarpov15: No longer necessary, removing this code.

expense/src/activities/payment.ts
vkarpov15: Do you recommend putting them in the same file? I don't have any strong feelings either way, but I think we should be consistent as to whether we recommend separate file per activity or all activities in 1 file.

vkarpov15:
@joebowbeer thanks for the reminder, fixed in badea4c

expense/src/workflows/expense.ts
```diff
@@ -0,0 +1,43 @@
...
+    status = 'TIMED_OUT';
+    return { status };
+  }
+
+  if (status === 'REJECTED') {
+    return { status };
+  }
+
+  await payment(expenseId);
+  status = 'COMPLETED';
```
bergundy, line 39: If you use my suggestion above, this turns into:

```suggestion
  if (status === 'APPROVED') {
    await payment(expenseId);
    status = 'COMPLETED';
  }
  return { status };
```

cancellation/src/workflows/example.ts
```diff
@@ -0,0 +1,26 @@
...
+  try {
+    await activityToBeCancelledWithTimeout();
+  } catch (_err) {
+    if (!(_err instanceof CancelledError)) {
+      throw _err;
+    }
+    err = _err;
+  }
+
+  return { message: err == null ? null : err.message };
```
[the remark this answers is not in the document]
vkarpov15: No reason worth noting, we can make this return `Promise<void>`.

expense/src/workflows/expense.ts
```diff
@@ -0,0 +1,43 @@
...
+    status = 'REJECTED';
+    signalTrigger.resolve();
+  }
+}
+
+async function main(expenseId: string): Promise<{ status: string }> {
+  await createExpense(expenseId);
+
+  if (status === 'CREATED') {
+    await Promise.race([signalTrigger, sleep(timeoutMS)]);
```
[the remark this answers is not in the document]
vkarpov15: I don't want to recommend doing that, because it might cause a race condition. `Promise.race()` doesn't cancel promises, so the `then(() => { status = 'TIMED_OUT' })` block would always execute. With the current implementation, that wouldn't be a problem, but you could see a case where that might lead to `status` getting updated in an unexpected way.

cancellation/package.json
```diff
@@ -1,12 +1,18 @@
...
+    "build.watch": "tsc --build --watch ./tsconfig.json",
+    "schedule-workflow": "node ./lib/starter",
+    "start": "node ./lib/worker"
   },
-  "name": "temporal-cancellation",
   "dependencies": {
     "temporalio": "0.3.x"
+  },
+  "devDependencies": {
+    "typescript": "^4.2.2",
```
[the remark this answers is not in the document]
vkarpov15: Done :+1:

cancellation/README.md
```diff
@@ -1,3 +1,8 @@
 # cancellation
 
-Doesn't work yet, core SDK doesn't implement Workflow cancellation
\ No newline at end of file
+## Steps to run this example
+
+1. Make sure the Temporal Server is running locally. Follow the [Quick install guide](https://docs.temporal.io/docs/server/quick-install) to do that.
+2. Run `npm install` to install dependencies
```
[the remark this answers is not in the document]
vkarpov15: Added `npm run build` and :eyes: on new project structure

expense/src/workflows/expense.ts
```diff
@@ -0,0 +1,43 @@
+import { createExpense } from '@activities/createExpense';
+import { payment } from '@activities/payment' ;
+import { Trigger, sleep } from '@temporalio/workflow';
+
+type expenseStatus = 'CREATED' | 'APPROVED' | 'REJECTED' | 'TIMED_OUT' | 'COMPLETED';
+let status: expenseStatus = 'CREATED';
+
+const signalTrigger = new Trigger<void>();
+const timeoutMS = 10000;
```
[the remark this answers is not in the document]
vkarpov15: @bergundy can you elaborate on this please?

cancellation/src/workflows/example.ts
```diff
@@ -0,0 +1,26 @@
...
+  startToCloseTimeout: '60s',
+  heartbeatTimeout: '3s',
+});
+
+async function main() {
+  let err;
+  try {
+    await activityToBeCancelledWithTimeout();
+  } catch (_err) {
+    if (!(_err instanceof CancelledError)) {
```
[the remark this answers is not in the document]
vkarpov15: I migrated the hello-world app to SDK 0.4 in this PR, but I'd rather create a separate PR after merging this one for upgrading the rest of the samples to 0.4.

expense/src/workflows/expense.ts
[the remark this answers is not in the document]
bergundy: Yeah, very good point, even though there's nothing async going on here until we `await payment()` below.
May I suggest you make the trigger type `Trigger<ExpenseStatus>` and resolve it with `accepted` or `rejected`.

```ts
status = await Promise.race([signalTrigger, sleep(timeoutMS).then(() => 'TIMED_OUT')]);
```

expense/src/starter/timeout.ts
```diff
@@ -0,0 +1,20 @@
...
+  const connection = new Connection();
+  const client = new WorkflowClient(connection.service);
+  const expense = client.stub<Expense>('expense', { taskQueue: 'tutorial' });
+
+  const expenseId = v4();
+
+  await expense.start(expenseId);
+
+  // Prints "Done: { status: 'TIMED_OUT' }" after approximately 10 seconds
+  console.log('Done:', await expense.result());
```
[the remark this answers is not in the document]
bergundy: Why not execute directly?

expense/src/workflows/expense.ts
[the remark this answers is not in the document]
bergundy: Refactor main below to:
```ts
async function main(expenseId: string, timeoutMs = 10000): ...
```

Then in  `expense/src/starter/timeout.ts`

Pass a shorter timeout value to demo the timeout.

cancellation/src/workflows/example.ts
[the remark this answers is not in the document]
bergundy: Sure, you might want to wait on that because I'm going to do some more breaking API changes this week.

expense/package.json
```diff
@@ -1,16 +1,25 @@
...
   "private": true,
+  "name": "temporal-expense",
   "scripts": {
-    "schedule-workflow": "node ./src/starter",
-    "server": "node ./src/server",
-    "start": "node ./src/worker"
+    "build": "tsc --build ./tsconfig.json",
+    "build.watch": "tsc --build --watch ./tsconfig.json",
+    "server": "node ./lib/server",
+    "start": "node ./lib/worker",
```
bergundy, line 9: Maybe call this script `worker`? Not critical.

expense/src/server/index.ts
```diff
@@ -1,21 +1,19 @@
...
-  completed: 'COMPLETED'
-})
+enum expenseState {
+  created = 'CREATED',
+  approved = 'APPROVED',
+  rejected = 'REJECTED',
+  completed = 'COMPLETED'
+}
 
 run().catch(err => console.log(err));
```
bergundy, line 10: You should exit the process with non 0 code in this case.
Also, best practice is to log to stderr.

bergundy (commented):
Good progress, keep it up

expense/src/workflows/expense.ts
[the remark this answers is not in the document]
vkarpov15: I like your suggestion, I think that's the right way to do this. Just implemented :+1:

expense/src/workflows/expense.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  if (status === ExpenseStatus.APPROVED) {
+    await payment(expenseId);
+    status = ExpenseStatus.COMPLETED;
+  }
+
+  return { status };
+}
+exports.workflow = { main, signals };
```
bergundy, line 37:
```suggestion
export const workflow: Expense = { main, signals };
```

expense/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "extends": "@tsconfig/node14/tsconfig.json",
+  "version": "4.2.2",
+  "compilerOptions": {
+    "sourceMap": true,
+    "esModuleInterop": true,
```
bergundy, line 6: This is already true when using `@tsconfig/node14`: https://www.npmjs.com/package/@tsconfig/node14

bergundy (commented):
@vkarpov15 please address my comments so we can merge this.

PR #5, opened 2021-09-05
Add mtls and opentelemetry samples

lorensr:


hello-world-mtls/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "private": true,
+  "scripts": {
+    "build": "tsc --build",
+    "build.watch": "tsc --build --watch",
+    "start": "ts-node src/worker.ts",
+    "start.watch": "nodemon src/worker.ts",
+    "workflow": "ts-node src/exec-workflow.ts"
+  },
+  "devDependencies": {
+    "@tsconfig/node16": "^1.0.0",
```
vkarpov15, line 12: Other examples use node 14. Do you think we should upgrade to node 16 across the board? Node 16 isn't quite the active LTS yet, but it will be as of October 26: https://nodejs.org/en/about/releases/

hello-world-mtls/src/exec-workflow.ts
```diff
@@ -0,0 +1,48 @@
+import fs from 'fs';
```
vkarpov15, line 1: I'd call this 'starter.ts' to be consistent with other examples and other docs.

vkarpov15 (approved):
Couple of minor comments but LGTM overall :+1:

hello-world-mtls/package.json
lorensr: Yeah, I think we're recommending 16 in general—Roey changed setup docs and nvmrc's to be 16.

hello-world-mtls/src/exec-workflow.ts
lorensr: I think `starter.ts` might be ambiguous to a newbie—just looking at the files, maybe it's starting a worker or activity:

image

What do you think about:

```
  "scripts": {
    ...
    "start-workflow": "ts-node src/start-workflow.ts"
  }
```

or 

```
  "scripts": {
    ...
    "execute-workflow": "ts-node src/execute-workflow.ts"
  }
```

These make it clear what's getting started/executed. The latter matches the function that's being called (`stub.execute`, not `stub.start`, at least in this example), and I think feels more in line with the fact that the script runs and completes, vs the `"start"` worker script that stays running until killed. Maybe we could always do the latter, or always use the verb that matches the stub function.

hello-world-mtls/package.json
bergundy: Let's switch to 16, it's close enough to LTS.

hello-world-mtls/src/exec-workflow.ts
bergundy: I don't know if we need a `start-workflow` script.
I kind of like calling the script `workflow` so it can be called with `npm run workflow`.

hello-world-mtls/package.json
bergundy: Thinking about it some more, it'd be nice to detect the user's node version when the script runs and use the correct tsconfig.

interceptors-opentelemetry/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "interceptors-opentelemetry-example",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "build": "tsc --build src/worker/tsconfig.json",
```
bergundy, line 6: Can we update these scripts to use nodemon?

interceptors-opentelemetry/src/worker/schedule-workflow.ts
```diff
@@ -0,0 +1,30 @@
+import { Connection, WorkflowClient } from '@temporalio/client';
```
bergundy, line 1: This file should probably be renamed to `exec-workflow.ts` or whatever we decide to use as a convention.

bergundy (approved):
Thanks!

Had a few minor comments.

hello-world-mtls/package.json
lorensr: Okay, so we can leave them as node16 here in case people manually clone, and have `@temporalio/create` replace it before `npm install`ing if user's node version is different.

hello-world-mtls/src/exec-workflow.ts
lorensr: I'm fine with that.

```
  "scripts": {
    ...
    "workflow": "ts-node src/execute-workflow.ts"
  }
```

interceptors-opentelemetry/package.json
lorensr: `hello-world-mtls` only uses it for `start.watch`—should it be used in others as well?
 
image

lorensr:
`interceptors-opentelemetry` isn't running—I'll fix it post React API

interceptors-opentelemetry/package.json
vkarpov15: Yeah I think adding start.watch to the other samples would be great

hello-world-mtls/package.json
lorensr: Implemented in https://github.com/lorensr/sdk-node/commit/3740f7728005d9f0cd6c18293788581b23944e59

interceptors-opentelemetry/package.json
bergundy: Agree

PR #6, opened 2021-09-09
update to Node SDK 0.4 across all examples

vkarpov15:
## Why?

Followup to #4, that PR was getting a bit too big so we're creating a separate PR to upgrade the new TS samples to SDK v0.4, taking advantage of the neat new `registerActivities()` syntax.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cancellation/src/workflows/example.ts
```diff
@@ -1,9 +1,10 @@
 'use strict';
 
-import { Context, CancelledError } from '@temporalio/workflow';
-import { activityToBeCancelled } from '@activities/activityToBeCancelled';
+import { CancelledFailure } from '@temporalio/common';
+import { Context } from '@temporalio/workflow';
+import * as activities from '../activities';
```
bergundy, line 5:
```suggestion
import type * as activities from '../activities';
```

cancellation/src/workflows/example.ts
```diff
@@ -12,9 +13,9 @@ const activityToBeCancelledWithTimeout = Context.configure(activityToBeCancelled
 async function main() {
   let err;
   try {
-    await activityToBeCancelledWithTimeout();
+    await activityToBeCancelled();
   } catch (_err) {
-    if (!(_err instanceof CancelledError)) {
+    if (!(_err instanceof CancelledFailure)) {
```
bergundy, line 18: This won't be `CancelledFailure`, use `isCancellation` instead: https://docs.temporal.io/docs/node/cancellation-scopes#cancelledfailure

expense/src/workflows/expense.ts
```diff
@@ -1,7 +1,12 @@
-import { ExpenseStatus } from '../interfaces/workflows';
-import { createExpense } from '@activities/createExpense';
-import { payment } from '@activities/payment' ;
+import { Expense, ExpenseStatus } from '../interfaces/workflows';
 import { Trigger, sleep } from '@temporalio/workflow';
+import { Context } from '@temporalio/workflow';
+import * as activities from '../activities';
```
bergundy, line 4: Same here and in the rest of the samples:
```suggestion
import type * as activities from '../activities';
```

bergundy (approved):
Had a few comments, I'm approving because I don't want to block but please address before merging.

Note that explicitly importing only the types is important here because it stresses the point that we don't import the implementations which likely cannot run in the Workflow isolate.

PR #8, opened 2021-09-13
update samples for new Reactlike API

sw-yx:
## What was changed

following through from https://github.com/temporalio/sdk-node/pull/225

## Why?

required for docs (https://github.com/temporalio/documentation/pull/627) and initializer to be up to sync too


hello-world/src/workflows.ts
```diff
@@ -0,0 +1,20 @@
...
+});
+
+/** A workflow that simply calls an activity */
+export const ExampleWorkflow = (name: string) => {
+  return {
+    async execute(): Promise<string> {
+      return await greet(name);
+    },
+  };
+};
```
bergundy, line 19: Do we like this syntax?

```suggestion
export const ExampleWorkflow = (name: string) => ({
  async execute(): Promise<string> {
    return await greet(name);
  },
});
```

bergundy (commented):
Are we getting rid of the interface completely?

I think we can for the hello world example but we should have a sample with the interface.

A few uses for the interface:
- Cross service / language invocation
- Make eslint happy when exporting the workflow

hello-world/src/workflows.ts
vkarpov15: I'm a big fan of this syntax too, more concise and less banana code.

hello-world/src/starter.ts
```diff
@@ -1,23 +1,22 @@
...
+  // via options passed the WorkflowClient constructor.
   const client = new WorkflowClient(connection.service);
-
-  // Create a typed client using the Example Workflow interface,
-  // Workflow will be started in the "default" namespace unless specified otherwise.
-  const example = client.stub<Example>('example', { taskQueue: 'tutorial' });
-
-  const result = await example.execute('Temporal');
+  // Create a typed client for the example Workflow.
+  const workflow = client.newWorkflowStub(ExampleWorkflow, { taskQueue: 'tutorial' });
```
vkarpov15, line 13: I think we settled on calling this `createWorkflowHandle()`

lorensr:
This this can be closed—was included in #9

vkarpov15:
I'm going to close this re: #9, feel free to reopen if I'm wrong there.

PR #9, opened 2021-09-15
Upgrade to SDK v0.4.2 and new React-style API

vkarpov15:
## What was changed

Address API changes for v0.4.2:

- `stub<Example>('example', opts)` -> `createWorkflowHandle(Example, opts)`
- `src/workflow/myname.ts` -> `src/workflow.ts`, `src/interfaces/workflow.ts` -> `src/interfaces.ts`
- `export const workflow = { main }` -> `export const workflowName = () => ({ async execute() { ... } })`
- `Context.registerActivities()` -> `createActivityHandle()`

## Why?

New API is a better fit with the "react for backend" metaphor, has more JS-friendly nomenclature, and makes it easier to test Workflows outside of isolates.

See #8

## Checklist

1. Closes 

2. How was this tested:

Went through README steps for each sample

3. Any docs updates needed?


expense/src/workflows.ts
```diff
@@ -0,0 +1,41 @@
+import { Expense, ExpenseStatus } from './interfaces';
+import { Trigger, sleep } from '@temporalio/workflow';
+import { createActivityHandle } from '@temporalio/workflow';
+import type * as activities from './activities';
+
+const { createExpense, payment } = createActivityHandle<typeof activities>({
+  scheduleToCloseTimeout: '5 minutes',
+});
+
+let status: ExpenseStatus = ExpenseStatus.CREATED;
```
bergundy, line 10: While this is totally fine and supported by the sdk I think we should discourage using mutable global state.
In the future we might want to run workflows in environments that are not isolated

bergundy (approved):
Thanks, lgtm with one comment

PR #10, opened 2021-09-16
add top-level eslint and prettier

vkarpov15:
## What was changed

Added a top-level `package.json` with ESLint and Prettier configs that are similar to the ones in the Node SDK. Ran both of those.

Also upgraded to node16 tsconfig across the board because I was getting a `File '@tsconfig/node16/tsconfig.json' not found` error with mixed versions.

## Why?

Basic checks to make sure that our samples don't have any obvious syntax errors and make sure that they abide by a reasonably consistent coding style.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cancellation/src/starter/index.ts
```diff
@@ -9,7 +9,7 @@ async function run() {
 
   await workflow.start();
 
-  await new Promise(resolve => setTimeout(resolve, 100));
+  await new Promise((resolve) => setTimeout(resolve, 100));
```
bergundy, line 12: In 2 years you'll be able to use this: https://nodejs.org/api/timers.html#timers_timerspromises_settimeout_delay_value_options

bergundy (approved):

cancellation/src/starter/index.ts
vkarpov15: That will be excellent, thanks for the heads up.

PR #11, opened 2021-09-16
add some basic test cases for http example

vkarpov15:
## What was changed

I'm working on adding some integration tests - I took a look and it looks like you should be able to stub out some activity logic using sinon. But I ran into an issue: when axios is stubbed out to return an error `sinon.stub(axios, 'get').callsFake(() => Promise.reject(new Error('42')));`, the `workflow.execute()` call hangs, and the error never bubbles up to the Workflow. Any ideas @bergundy ?

I borrowed some test patterns from here: https://github.com/temporalio/sdk-node/blob/main/packages/test/src/test-integration.ts

## Why?

Right now we don't have a full testing solution that handles stubbing activities and advancing time, but I wanted to see if I could use sinon to get a scrappy approach working.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/http/test/workflow.test.ts
+++ b/http/test/workflow.test.ts
@@ -1,0 +1,52 @@
+import { Connection, WorkflowClient } from '@temporalio/client';
+import { Worker, DefaultLogger } from '@temporalio/worker';
+import { describe, before, after, it } from 'mocha';
+import { example } from '../lib/workflows';
+import assert from 'assert';
+import axios from 'axios';
+import sinon from 'sinon';
+
+describe('example workflow', function() {
+  let worker = null;
+  let runPromise = null;
+
+  before(async function() {
+    this.timeout(10000);
+    worker = await Worker.create({
+      workDir: `${__dirname}/../lib`,
+      taskQueue: 'tutorial20210916',
+      logger: new DefaultLogger('ERROR'),
+    });
+
+    runPromise = worker.run();
+  });
+
+  after(async function() {
+    worker.shutdown();
+    await runPromise;
+  });
+
+  it('returns correct result', async function() {
+    const connection = new Connection();
+    const client = new WorkflowClient(connection.service);
+
+    const workflow = client.createWorkflowHandle(example, { taskQueue: 'tutorial20210916' });
+
+    const result = await workflow.execute();
+    assert.equal(result, 'The answer is 42');
+  });
+
+  it('handles errors', async function() {
+    this.timeout(5000);
+    const connection = new Connection();
+    const client = new WorkflowClient(connection.service);
+
+    const workflow = client.createWorkflowHandle(example, { taskQueue: 'tutorial20210916' });
+
+    //sinon.stub(axios, 'get').callsFake(() => Promise.resolve({ data: { args: { answer: 42 } } }));
+    sinon.stub(axios, 'get').callsFake(() => Promise.reject(new Error('42')));
+
+    // This call hangs:
+    const result = await workflow.execute();
+  });
+});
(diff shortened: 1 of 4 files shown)


http/test/workflow.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    const connection = new Connection();
+    const client = new WorkflowClient(connection.service);
+
+    const workflow = client.createWorkflowHandle(example, { taskQueue: 'tutorial20210916' });
+
+    //sinon.stub(axios, 'get').callsFake(() => Promise.resolve({ data: { args: { answer: 42 } } }));
+    sinon.stub(axios, 'get').callsFake(() => Promise.reject(new Error('42')));
+
+    // This call hangs:
+    const result = await workflow.execute();
```
bergundy, line 50: My best bet here is that you have unlimited retries for the activity.
If you want the activity to eventually complete with failure change the retry policy: https://github.com/temporalio/sdk-node/blob/23da8507a90049774cc6ddd28a8ec3dfc37d345f/packages/test/src/workflows/activity-failure.ts#L8

bergundy (approved):
lgtm

http/test/workflow.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    workflow = client.createWorkflowHandle(example, { taskQueue: 'testhttp' });
+  });
+
+  after(async function() {
+    worker.shutdown();
+    await runPromise;
+  });
+
+  afterEach(() => {
+    axios.get.restore && axios.get.restore();
```
joebowbeer, line 38:
```suggestion
    sinon.restore();
```

I think it's simpler and more maintainable to `sinon.restore()` in afterEach.

joebowbeer (commented):
@vkarpov15 Suggest using `sinon.restore()` in simple tests like these.

http/test/workflow.test.ts
vkarpov15: Great suggestion, thanks!

http/test/workflow.test.ts
```diff
@@ -0,0 +1,67 @@
+import { Connection, WorkflowClient } from '@temporalio/client';
+import { Worker, DefaultLogger } from '@temporalio/worker';
+import { describe, before, after, afterEach, it } from 'mocha';
+import { example } from '../lib/workflows';
```
joebowbeer, line 4: Can these duplications of the outDir path be removed?

I would prefer that outDir only needed to be changed in one place: tsconfig.json

http/test/workflow.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+describe('example workflow', function() {
+  let runPromise = null;
+  let worker = null;
+  let workflow = null;
+
+  before(async function() {
+    this.timeout(10000);
+    worker = await Worker.create({
+      workDir: `${__dirname}/../lib`,
```
joebowbeer, line 17: outDir duplicated here as well.

joebowbeer (commented):
Commenting on undesirable duplication of outDir

http/test/workflow.test.ts
bergundy: This is unnecessary, you can import directly from `./workflows`

http/test/workflow.test.ts
bergundy: Should be `workDir: __dirname`

http/test/workflow.test.ts
joebowbeer: I think this is a consequence of using ts-mocha to run the tests from the test folder, while the code under test is precompiled into lib.

PR #12, opened 2021-09-17
Use pre-commit hook and standardize files

lorensr:
Standardize `package.json`s

## What was changed

Use symbolic links for sample files (https://github.com/temporalio/sdk-node/issues/136):
- tsconfig.json
- .nvmrc
- .gitignore

## Why?

So that when we update, we only have to edit one file.

## Checklist

2. How was this tested:

They don't run, but they didn't run for me before either (webpack error). Will fix after the new `workDir` API is released.


expense/package.json
```diff
@@ -1,14 +1,24 @@
...
-    "server": "node ./lib/server",
-    "starter-approve": "node ./lib/starter/approve",
-    "starter-timeout": "node ./lib/starter/timeout",
-    "worker": "node ./lib/worker"
+    "build": "tsc --build",
+    "build.watch": "tsc --build --watch",
+    "server": "ts-node src/server",
+    "start": "ts-node src/worker.ts",
+    "start.watch": "nodemon src/worker.ts",
+    "workflow-approve": "ts-node src/execute/approve",
```
vkarpov15, line 11: Good rename, thanks :+1:

expense/package.json
```diff
@@ -1,14 +1,24 @@
...
+    "workflow-timeout": "ts-node src/execute/timeout"
+  },
+  "nodemonConfig": {
```
vkarpov15, line 14: Maybe we do a shared [`nodemon.json` file](https://github.com/remy/nodemon#config-files) as well?

hello-world/src/worker.ts
```diff
@@ -13,10 +15,12 @@ async function run() {
   // In order to configure the server connection parameters and other global options,
   // use the Core.install() method to configure the Rust Core SDK singleton.
   const worker = await Worker.create({
-    workDir: __dirname,
-    taskQueue: 'tutorial'
+    workflowsPath: path.join(__dirname, 'workflows'),
```
vkarpov15, line 18: Why is this change necessary? I think we generally encourage people to use `workDir`...

vkarpov15 (commented):
Nice, I like this overall. My only concern is that the shared config adds friction for devs that want to copy one of the samples into a separate repo using `cp -r`. The friction isn't too bad, might be worth adding a note to the README though.

hello-world/src/worker.ts
lorensr: Planning on removing workDir, replacing with something like:

image

vkarpov15: Cool, I missed that convo. Thanks for pointing that out :+1:

joebowbeer:
I suggest fixing the samples before addressing this. It is an easy fix.

PR #14, opened 2021-09-21
fix: import from workflows index

### LINKED ISSUE #13: [Bug] Need to move workflows.ts to index.ts
### Describe the bug

temporalio 0.4.2 will not automatically (workDir) import from workflows.ts - it expects workflows/index.ts

### To Reproduce

1. Go to hello-world
2. Clean checkout and build using node 16
3. Follow steps in README

Similar with http. Probably others as well, but I didn't look or try.

### Expected behavior

Samples work.

### Versions

- OS: Mac
- Temporal Version 1.12.1

### Additional Context

The underlying problem is described in https://github.com/temporalio/sdk-node/issues/258

joebowbeer:
## What was changed

Fixes #13 (hello-world, http)

## Checklist

1. Closes #13 
2. How this was tested: Followed steps in README

### DIFF
--- a/http/tsconfig.json
+++ b/http/tsconfig.json
@@ -6,6 +6,11 @@
     "rootDir": "./src",
     "outDir": "./lib"
   },
-  "include": ["./src/**/*.ts"],
-  "exclude": ["./node_modules/*"]
+  "include": [
+    "src/**/*.ts"
+  ],
+  "exclude": [
+    "node_modules",
+    "test"
+  ]
 }
--- a/hello-world/package.json
+++ b/hello-world/package.json
@@ -12,7 +12,7 @@
     "temporalio": "0.4.2"
   },
   "devDependencies": {
-    "typescript": "^4.4.2",
-    "@tsconfig/node16": "^1.0.0"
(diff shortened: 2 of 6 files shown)


http/tsconfig.json
```diff
@@ -6,6 +6,11 @@
...
   },
-  "include": ["./src/**/*.ts"],
-  "exclude": ["./node_modules/*"]
+  "include": [
+    "src/**/*.ts"
+  ],
+  "exclude": [
+    "node_modules",
+    "test"
+  ]
```
joebowbeer, line 15: This is gratuitous, but I think it's a little clearer

vkarpov15:
I hate to say this, but this sample works fine on my machine regardless of whether I use `workflows.ts` or `workflows/index.ts`. The issue described by https://github.com/temporalio/sdk-node/issues/258 specifically refers to `workflowsPath`, which we don't use in this example, so maybe that issue only affects `workflowsPath`?

What error do you get when running this sample as currently written?

joebowbeer:
Closing as obsolete

PR #16, opened 2021-09-23
rename `workflow` -> `handle` in starter / `execute-workflow.ts`

vkarpov15:
## What was changed

As per discussion in Slack, `workflow` is renamed to `handle` to be more consistent with `createWorkflowHandle()`

## Why?

Better nomenclature. `createWorkflowHandle()` returns a workflow handle, so better to call the variable `workflowHandle` or `handle` than `workflow`.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/hello-world-mtls/package.json
+++ b/hello-world-mtls/package.json
@@ -10,7 +10,7 @@
     "workflow": "ts-node src/execute-workflow.ts"
   },
   "dependencies": {
-    "temporalio": "^0.4.1"
+    "temporalio": "0.4.1"
   },
   "devDependencies": {
     "@tsconfig/node16": "^1.0.0",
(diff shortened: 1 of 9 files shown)


bergundy (approved):

sw-yx:
oh! lol i wasnt aware that we definitively decided on this one heheh. i'm down

hello-world-mtls/package.json
```diff
@@ -10,7 +10,7 @@
     "workflow": "ts-node src/execute-workflow.ts"
   },
   "dependencies": {
-    "temporalio": "^0.4.1"
+    "temporalio": "0.4.1"
```
sw-yx, line 13: why pin?

vkarpov15: Because hello-world-mtls is still written using v0.4.1 syntax, it won't compile with v0.4.2.

PR #17, opened 2021-09-23
add timer sample

vkarpov15:
## What was changed

Added a sample that's analogous to the `timer` sample in samples-go: https://github.com/temporalio/samples-go/tree/master/timer

## Why?

While similar to the `expense` sample, this sample has a couple of noteworthy differences that I think make it a useful example:

1) Provides an example of integrating with a real API (Mailgun)
2) Provides an example of using environment variables and dotenv to configure activities
3) Example of using `Promise.race()` and `sleep()` to create a soft timeout for an activity, as opposed to waiting on a signal with a trigger

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/timer/README.md
+++ b/timer/README.md
@@ -1,3 +1,7 @@
+# Timer Reminder Example
+
+This example shows how to use a Timer to send a reminder email after some time elapses.
+
 ## Steps to run this example
 
 1. Make sure the Temporal Server is running locally. Follow the [Quick install guide](https://docs.temporal.io/docs/server/quick-install) to do that.
--- a/timer/README.md
+++ b/timer/README.md
@@ -1,0 +1,9 @@
+## Steps to run this example
+
+1. Make sure the Temporal Server is running locally. Follow the [Quick install guide](https://docs.temporal.io/docs/server/quick-install) to do that.
+2. Run `npm install` to install dependencies.
+3. Set up an account with [Mailgun](https://www.mailgun.com/)
+4. Create a `.env` file with the following environment variables: `MAILGUN_API`, `MAILGUN_DOMAIN`, and `ADMIN_EMAIL`
+5. Run `npm run build` to compile the project.
+6. Run `npm run worker` to start the worker. Leave the worker process running.
+7. Run `npm run workflow-slow` to start the example. Should print out "Done" and send an email to the email specified in the `ADMIN_EMAIL` environment variable
--- a/timer/src/activities.ts
+++ b/timer/src/activities.ts
@@ -1,0 +1,25 @@
+import { config } from 'dotenv';
+import mailgun from 'mailgun-js';
+
+config();
+
+const apiKey: string = process.env.MAILGUN_API as string;
+const domain: string = process.env.MAILGUN_DOMAIN as string;
+const to: string = process.env.ADMIN_EMAIL as string;
+
+const mg = mailgun({ apiKey, domain });
+
+export async function processOrder(sleepMS = 1000): Promise<void> {
+  await new Promise(resolve => setTimeout(resolve, sleepMS));
+};
+
+export async function sendNotificationEmail(): Promise<void> {
+  const data = {
+    to,
+    subject: 'Order processing taking longer than expected',
+    from: 'Temporal Bot <temporal@' + process.env.MAILGUN_DOMAIN + '>',
+    html: 'Order processing is taking longer than expected, but don\'t worry, the job is still running!'
+  };
+
+  await mg.messages().send(data);
+};
--- a/timer/src/activities.ts
+++ b/timer/src/activities.ts
@@ -1,3 +1,4 @@
+// @@@SNIPSTART nodejs-timer-reminder-activity
 import { config } from 'dotenv';
 import mailgun from 'mailgun-js';
 
@@ -23,3 +24,4 @@ export async function sendNotificationEmail(): Promise<void> {
 
   await mg.messages().send(data);
 };
+// @@@SNIPEND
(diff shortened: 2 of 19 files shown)


timer/src/activities.ts
```diff
@@ -0,0 +1,25 @@
+import { config } from 'dotenv';
+import mailgun from 'mailgun-js';
+
+config();
```
bergundy, line 4: I would put this all in the worker where all initialization happens.
See my other pr for injecting the client into the activities

timer/src/activities.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const apiKey: string = process.env.MAILGUN_API as string;
+const domain: string = process.env.MAILGUN_DOMAIN as string;
+const to: string = process.env.ADMIN_EMAIL as string;
+
+const mg = mailgun({ apiKey, domain });
+
+export async function processOrder(sleepMS = 1000): Promise<void> {
+  await new Promise(resolve => setTimeout(resolve, sleepMS));
+};
```
bergundy, line 14: Use Context.current().sleep() which throws if the activity is cancelled

timer/src/workflows.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ProcessOrder } from './interfaces';
+// Only import the activity types
+import type * as activities from './activities';
+
+const { processOrder, sendNotificationEmail } = createActivityHandle<typeof activities>({
+  startToCloseTimeout: '5 minutes',
+});
+
+/** A workflow that simply calls an activity */
+export const processOrderWorkflow: ProcessOrder = (orderProcessingMS = 1000, sendDelayedEmailTimeoutMS = 100) => ({
```
bergundy, line 11: Better to take an object argument for readability

bergundy (changes requested):

timer/src/activities.ts
```diff
@@ -0,0 +1,25 @@
+import { config } from 'dotenv';
```
sw-yx, line 1:
```suggestion
// @@@SNIPSTART nodejs-timer-activity
import { config } from 'dotenv';
```

timer/src/activities.ts
```diff
@@ -0,0 +1,25 @@
...
+export async function sendNotificationEmail(): Promise<void> {
+  const data = {
+    to,
+    subject: 'Order processing taking longer than expected',
+    from: 'Temporal Bot <temporal@' + process.env.MAILGUN_DOMAIN + '>',
+    html: 'Order processing is taking longer than expected, but don\'t worry, the job is still running!'
+  };
+
+  await mg.messages().send(data);
+};
```
sw-yx, line 25:
```suggestion
};
// @@@SNIPEND
```

timer/src/workflows.ts
```diff
@@ -0,0 +1,24 @@
+import { createActivityHandle, sleep } from '@temporalio/workflow';
```
sw-yx, line 1:
```suggestion
// @@@SNIPSTART nodejs-timer-reminder-workflow
import { createActivityHandle, sleep } from '@temporalio/workflow';
```

timer/src/workflows.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    await Promise.race([processOrderPromise, sleep(sendDelayedEmailTimeoutMS)]);
+
+    if (processing) {
+      await sendNotificationEmail();
+
+      await processOrderPromise;
+    }
+  }
+});
```
sw-yx, line 24:
```suggestion
});
// @@@SNIPEND
```

timer/README.md
```diff
@@ -0,0 +1,9 @@
+## Steps to run this example
+
```
sw-yx, line 2: just add some description to give the reader some idea of what the heck is going on?

```suggestion
# Timer Reminder Example

This example shows how to use a Timer to send a reminder email after some time elapses.

## Steps to run this example

```

timer/src/activities.ts
vkarpov15: Is the Worker where we want to formally recommend initializing config, etc.? That approach works for me, my only potential thought is that it might be a bit confusing because`worker.ts` currently doesn't explicitly access `activities.ts`, which may be a tad confusing.

bergundy: I’d say yes because the way the code is written today requires stubbing out a bunch of different things to make it testable

timer/src/activities.ts
```diff
@@ -0,0 +1,27 @@
+// @@@SNIPSTART nodejs-timer-reminder-activity
+import { config } from 'dotenv';
+import mailgun from 'mailgun-js';
+
+config();
+
+const apiKey: string = process.env.MAILGUN_API as string;
```
bergundy, line 7: This cast is wrong, the type is string | undefined.
Instead of casting make a runtime assertion.

vkarpov15: Done using elvis operator, let me know if you prefer something else for backwards compat.

sw-yx (approved):

sw-yx:
nice work!

timer/src/activities.ts
```diff
@@ -0,0 +1,25 @@
+// @@@SNIPSTART nodejs-timer-reminder-activity
+import { Context } from '@temporalio/activity';
+import mailgun from 'mailgun-js';
+
+const apiKey: string = process.env.MAILGUN_API ?? '';
+const domain: string = process.env.MAILGUN_DOMAIN ?? '';
+const to: string = process.env.ADMIN_EMAIL ?? '';
```
bergundy, line 7: I would still really like this sample to direct devs to use DI for injecting clients into activities instead of initializing them in the top level for better testability.
One other issue with these lines is that there's no assertion that all required env vars are provided, this is not good practice, prefer to fail fast.

PR #18, opened 2021-09-23
Add Activity dependency injection sample

bergundy:


sw-yx (approved):
i hope you dont mind i tweaked your examples a bit as i believe that DI only makes sense when we are sharing it between _multiple_ activities

activities-dependency-injection/test/workflow.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    workflow = client.createWorkflowHandle(example, { taskQueue: 'testhttp' });
+  });
+
+  after(async function() {
+    worker.shutdown();
+    await runPromise;
+  });
+
+  afterEach(() => {
+    axios.get.restore && axios.get.restore();
```
vkarpov15, line 38: Use `sinon.restore()` instead, although I don't think this test is supposed to be here because it looks like it was copied as-is from the http example

activities-dependency-injection/src/worker.ts
```diff
@@ -0,0 +1,23 @@
+import { Worker } from '@temporalio/worker';
+
+// @@@SNIPSTART typescript-activity-deps-worker
+import { createActivities } from './activities';
+
+async function run() {
+  // mock db connection initialization in Worker
+  const dbConnection = async () => 'Temporal'
```
vkarpov15, line 8: Would you create a database connection in a Workflow? Most database drivers I've worked with open a persistent socket connection when you connect to the database, so that sounds like the sort of side-effect thing that wouldn't work well in a Workflow.

bergundy: I don’t understand your comment, this is the worker

bergundy:
> i hope you dont mind i tweaked your examples a bit as i believe that DI only makes sense when we are sharing it between multiple activities
No problem

activities-dependency-injection/test/workflow.test.ts
bergundy: Writing code at midnight is probably not the best idea.

PR #20, opened 2021-09-29
Use npm-run-all

lorensr:
In nextjs sample:
- Simplify `npm run dev` command, and allow user to stop all commands with Cmd-C (currently you have to `kill <pid>`)
- It appears from my cursory research that `command:scope` is more common for script naming. See for instance these examples:
https://stackoverflow.com/questions/47606101/what-is-colon-in-npm-script-names
https://github.com/mysticatea/npm-run-all/

@sw-yx


sw-yx:
thank you :)

PR #25, opened 2021-10-09
Fix prettier & linting; improve testing

lorensr:
Done:

* Add CI
* Enable tests to run without building
* Run prettier
* Fix some of the linting errors

Any tips on fixing the rest of the lint errors pictured below?

- Do we want to silence the missing function return type?
- For the `workflow.test.js`, add to `activites-examples/tsconfig` (can add to `include`, but then have to change `rootDir`) or create a new `test/tsconfig`?
- What to do for all the JS files?

(Feel free to add commits to this PR if you'd like)

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
           # Comment out next line when testing with `act`
-          # cache: 'npm'
+          cache: 'npm'
       - run: npm install
       - run: npx prettier . --check
       # - run: npm run lint


.gitignore
```diff
@@ -103,6 +103,7 @@ dist
 # TernJS port file
 .tern-port
 
-package-lock.json
+*/package-lock.json
+*/**/package-lock.json
```
bergundy, line 107: I'm pretty sure that `package-lock.json` is enough here.

.prettierignore
```diff
@@ -0,0 +1,2 @@
+.prettierignore
```
bergundy, line 1: You'll probably want to ignore generated JS files as well, see the list in the sdk-node repo: https://github.com/temporalio/sdk-node/blob/main/.prettierignore

activities-dependency-injection/README.md
```diff
@@ -7,7 +7,7 @@ This sample shows how to share dependencies between activities: for example, whe
 1. Make sure the Temporal Server is running locally. Follow the [Quick install guide](https://docs.temporal.io/docs/server/quick-install) to do that.
 2. Run `npm install` to install dependencies.
 3. Run `npm run build` to compile the project.
-3. Run `npm start.watch` to start the worker and reload on changes. Leave the worker process running.
+4. Run `npm start.watch` to start the worker and reload on changes. Leave the worker process running.
```
bergundy, line 10: This would have been solved by just using `1.` for all the list items.

activities-dependency-injection/src/activities.ts
```diff
@@ -1,11 +1,11 @@
 // @@@SNIPSTART typescript-activity-with-deps
-export const createActivities = (dbConnection: Function) => ({
+export const createActivities = (dbConnection: () => string) => ({
```
bergundy, line 2: I fixed this in #26

bergundy (commented):
Overall looks like a good start, we should test the actual samples too somehow.

activities-dependency-injection/README.md
lorensr: Think this was autofixed by prettier

hello-world/package.json
```diff
@@ -19,10 +20,13 @@
     ]
   },
   "dependencies": {
-    "temporalio": "0.5.0"
+    "temporalio": "0.4.2"
```
vkarpov15, line 23: Why is this change necessary? Might conflict with #29

package.json
```diff
@@ -2,6 +2,8 @@
   "name": "samples-node",
   "private": true,
   "scripts": {
+    "pretest": "cd activities-examples && npm install",
+    "test": "cd activities-examples && npm test",
```
vkarpov15, line 6: Maybe use [pushd and popd](https://fishshell.com/docs/current/cmds/pushd.html) instead so this doesn't change the user's cwd? Also, would be great if we didn't have to explicitly list all examples here. Maybe we write a test script instead?

package.json
```diff
@@ -15,6 +17,7 @@
     "eslint-config-prettier": "^8.3.0",
     "husky": "^7.0.2",
     "lint-staged": "^11.2.3",
+    "next": "^11.1.2",
```
vkarpov15, line 20: Is this necessary? I don't think we use next outside of the nextjs examples.

package.json
lorensr: Will try switching to lerna

hello-world/package.json
lorensr: I don't remember. It did, and it's now 0.11.x 😄

.shared/.eslintignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+lib
+.eslintrc.js
```
bergundy, line 3: We can just lint this once and not ignore it

.shared/tsconfig.json
```diff
@@ -2,8 +2,6 @@
   "extends": "@tsconfig/node16/tsconfig.json",
   "version": "4.4.2",
   "compilerOptions": {
-    "emitDecoratorMetadata": false,
-    "experimentalDecorators": false,
```
bergundy, line 6: There's no way to comment on this but we don't need `composite`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on:
+  pull_request:
+    branches: [$default-branch]
+  push:
+    branches: [$default-branch, next]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
bergundy, line 11: when we do actual testing let's add mac and later windows (missing SDK support)

.eslintrc.json
```diff
@@ -10,5 +10,8 @@
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
     "prettier"
-  ]
+  ],
+  "rules": {
```
bergundy, line 14: Why do we need global `eslint` and `eslintignore`?

activities-dependency-injection/package.json
```diff
@@ -7,7 +7,8 @@
     "build.watch": "tsc --build --watch",
     "start": "ts-node src/worker.ts",
     "start.watch": "nodemon src/worker.ts",
-    "workflow": "ts-node src/execute-workflow.ts"
+    "workflow": "ts-node src/execute-workflow.ts",
+    "lint": "eslint ."
```
bergundy, line 11: I wish we had a template for this too

activities-dependency-injection/src/worker.ts
```diff
@@ -6,6 +6,7 @@ import { createActivities } from './activities';
 async function run() {
   // Mock DB connection initialization in Worker
   const db = {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
bergundy, line 9: Instead please add the following to the shared eslintrc:
```
      "@typescript-eslint/no-unused-vars": [
        1,                                                                                                                                                                                                          {
          "argsIgnorePattern": "^_"                                                                                                                                                                                 }
      ]
```

activities-examples/src/activities/fake-progress.ts
```diff
@@ -0,0 +1,19 @@
+// @@@SNIPSTART nodejs-activity-fake-progress
```
bergundy, line 1: I deleted this file, should not be here

activities-examples/src/test/workflow.test.ts
```diff
@@ -67,7 +71,11 @@ describe('example workflow', function () {
 
     await assert.rejects(
       () => workflow.execute(),
-      (err: any) => err.name === 'WorkflowExecutionFailedError' && err.cause.cause.message === 'example error'
+      (err: unknown) =>
+        err instanceof WorkflowExecutionFailedError &&
+        err.cause &&
+        'cause' in err.cause &&
+        err.cause['cause']['message'] === 'example error'
```
bergundy, line 78: If we want to be exact use something like this:
`err.cause instanceof ActivityFailure && err.cause.cause && intstanceof ApplicationFailure && err.cause.cause.message === 'example error'`

ecommerce-one-click/nextjs/temporal/src/workflows/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
bergundy, line 1: Let's not ignore the return type here, only in `createActivites`

bergundy: OIC, this is outdated

ecommerce-one-click/nextjs/tsconfig.json
```diff
@@ -18,12 +14,6 @@
...
   },
-  "include": [
-    "next-env.d.ts",
-    "**/*.ts",
-    "**/*.tsx"
-  ],
-  "exclude": [
-    "node_modules"
-  ]
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
```
bergundy, line 17: This could be a composite project with references to the worker so it's all built together.
Not for this PR though.

.shared/.eslintrc.js
```diff
@@ -0,0 +1,15 @@
...
+    project: './tsconfig.json',
+    tsconfigRootDir: __dirname,
+  },
+  plugins: ['@typescript-eslint'],
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/eslint-recommended',
+    'plugin:@typescript-eslint/recommended',
+    'prettier',
+  ],
```
bergundy, line 14: Please add these too:
```
      "object-shorthand": ["error", "always"],
      "deprecation/deprecation": "warn",
      "@typescript-eslint/no-explicit-any": 0,
```

.eslintrc.json
bergundy: In `sdk-node` we have it because we don't have the per package config.
Here, I think it'd make sense to actually run eslint per-package, otherwise what's the point of having those configs?

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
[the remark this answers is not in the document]
lorensr: Think this is included by default by husky? At least I didn't add it

image

.husky/pre-commit
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
lorensr: Looks like I can make it a node script instead

activities-dependency-injection/package.json
lorensr: For package.json? I think too many differences to be worthwhile

.eslintrc.json
bergundy: I know, my point is that we should lint with the per-package config so users aren't surprised if they try to lint our code and and it fails.

activities-dependency-injection/package.json
bergundy: I didn't mean the whole file but it would have been nice to have some shared structure that we can sync.
Definitely not for this PR.

package.json
bergundy: pushd and popd don't matter here, script runs in a different shell

package.json
bergundy: Basically I think we should run lint per sample (concurrently with lerna)

.husky/.gitignore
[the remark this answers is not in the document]
bergundy: oh, only used husky once, didn't know.

.shared/.eslintignore
bergundy: oh, okay, I see why now.

.husky/pre-commit
```diff
@@ -0,0 +1,31 @@
...
+# . "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
+
+for file in *; do
+  if [[ -d "$file" ]] # is a directory?
+  then
+    dir=$file
+    if ! [[ " node_modules ecommerce-one-click " =~ " $dir " ]]
```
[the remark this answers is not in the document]
bergundy: I'd at least just prompt and say something like:
```
Running pre-commit hook, this will overwrite any config files created in samples, proceed? [Y/n]
```

.husky/pre-commit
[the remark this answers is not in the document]
bergundy: Please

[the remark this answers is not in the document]
bergundy: want to try [zx](https://github.com/google/zx)? I haven't tried it yet.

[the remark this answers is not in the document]
lorensr: In order to change `#!/bin/sh` to `#!/usr/bin/env zx`, I think all devs would need to have zx globally installed. At least it didn't work for me with it only in `node_modules`. Dunno if that would be worth it?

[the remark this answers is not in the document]
lorensr: Turned off verbosity. Output now:

```
$ npx zx .shared/copy-shared-files.mjs 
Running pre-commit hook.
This will overwrite any changes made to config files in samples (like hello-world/tsconfig.json).
Proceed? [Y/n] y
Copying config files from .shared/ to samples... done.
```

lorensr:
Fixed otel with `npm update temporalio` and updating `workflowModules`.

.shared/copy-shared-files.mjs
```diff
@@ -0,0 +1,65 @@
+const NON_SAMPLES = ['node_modules'];
```
bergundy, line 1: nit: Should this be in `.shared`? Maybe create a scripts directory?

PR #29, opened 2021-10-13
Update to wf api r3

bergundy:
Best reviewed commit by commit

### DIFF
--- a/timer/src/workflows.ts
+++ b/timer/src/workflows.ts
@@ -12,7 +12,6 @@ export interface ProcessOrderOptions {
   sendDelayedEmailTimeoutMS: number;
 }
 
-/** A workflow that simply calls an activity */
 export async function processOrderWorkflow({
   orderProcessingMS,
   sendDelayedEmailTimeoutMS,


expense/src/execute/approve.ts
```diff
@@ -12,7 +12,7 @@ async function run() {
   await handle.start(expenseId);
 
   await new Promise((resolve) => setTimeout(resolve, 50));
-  await handle.signal.approve();
+  await handle.signal(approveSignal);
```
joebowbeer, line 15: If the workflow does not support this signal, type-checking won't catch it here, right? Will it be caught at runtime?

sw-yx: no, it will not be caught at runtime, and the signal will just silently be recorded with no listener. Roey and I have considered printing a warning when signals are sent with no listener, but i am unclear if this should be implemented at the SDK or Server level.

expense/src/server/index.ts
```diff
@@ -1,22 +1,38 @@
...
 
 async function run() {
   const app = express();
   app.use(express.json());
 
-  const allExpenses = new Map();
+  const allExpenses = new Map<string, ExpenseStatus>();
 
-  app.get('/', function (req, res) {
+  app.get('/', function (_req, res) {
```
vkarpov15, line 31: Why `_req` here? So TS doesn't complain about unused vars?

PR #32, opened 2021-10-16
Add sticky activity worker example

bergundy:


vkarpov15 (approved):
Would be great to link to docs about sticky activities but otherwise LGTM :+1:

PR #34, opened 2021-10-18
add cron workflows sample

swyxio:
## What was changed
add cron workflows sample

image

image


cron-workflows/src/activities.ts
```diff
@@ -0,0 +1,5 @@
+export async function greet(name: string, wfTime: string): Promise<void> {
```
bergundy, line 1: Should we rename this to something other than `greet`?

bergundy (approved):

PR #37, opened 2021-10-18
wip

bergundy:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


query-subscriptions/src/subscribe.ts
```diff
@@ -0,0 +1,101 @@
...
+  readBlockTimeMs?: number;
+}
+
+/**
+ * Helper for polling on versioned patches as generated by the subscriptions interceptors.
+ */
+async function poll(redis: Redis.Redis, workflowId: string, version: number, opts: PollOptions = {}) {
+  const { maxEntries, readBlockTimeMs } = opts;
+
+  for (;;) {
```
sw-yx, line 36: unusual way to while(true)

PR #43, opened 2021-10-29
Add message to display after @temporalio/create

lorensr:
- Move script from .shared/ to .scripts/
- Remove mention of old ecommerce/ dir
- Add .shared/.post-create
- Update non-shared post create messages


lorensr:
Run by `chalk -t "$(cat timer/.post-create)"`. I'll update `@temporalio/create` next.

sw-yx (approved):
ok but lets not overengineer install experience beyond this ok? think we reached diminishing returns here

bergundy (approved):

PR #47, opened 2021-11-01
Run npm test in CI

lorensr:
## What was changed

Got Temporal Server running in GH action so that it can run `npm test`.

image


.github/workflows/ci.yml
```diff
@@ -20,9 +20,25 @@ jobs:
...
-          # Comment out next line when testing with `act`
+          # Comment out cache line when testing with act:
+          # act --platform ubuntu-latest=lucasalt/act_base:latest
           cache: 'npm'
       - run: npm install
       - run: npx prettier . --check
       # - run: npm run lint
-      # - run: npm test
+      - name: Get Temporal docker-compose.yml
+        run: wget https://raw.githubusercontent.com/temporalio/docker-compose/v1.13.0/docker-compose.yml
```
bergundy, line 30: I know we set a specific version in `sdk-typescript` but let's try and be more loose here and use `main` so we automatically update when a new server version is released.

bergundy (approved):
I'm approving this but I think we should have test coverage for running workers + starters as well.
Those might break unexpectedly.
Not blocking this PR though.

PR #51, opened 2021-11-03
Add production sample with prebuilt workflow bundle

lorensr:
Wasn't sure where to put the bundle. Chose `/workflow-bundle.js`. Didn't like it in `lib/` because it got overwritten on `npm build`.

## Why?

So we have a full working sample we can point to from the new deployment docs: https://github.com/temporalio/documentation/pull/719


production/scripts/build-workflow-bundle.ts
```diff
@@ -0,0 +1,15 @@
...
+import path from 'path';
+
+async function bundle() {
+  const { code } = await bundleWorkflowCode({
+    workflowsPath: require.resolve('../src/workflows'),
+  });
+
+  await writeFile(path.join(__dirname, '../workflow-bundle.js'), code);
+
+  console.log('Bundle written to workflow-bundle.js');
```
bergundy, line 12: nit: might as well print the entire path here.

production/src/worker.ts
```diff
@@ -0,0 +1,19 @@
+import { Worker } from '@temporalio/worker';
+import * as activities from './activities';
+
+async function run() {
+  const worker = await Worker.create({
+    ...(process.env.NODE_ENV === 'production'
```
bergundy, line 6: nit: Maybe extract this to a function?

bergundy (approved):
Just the snippets need to be removed, nothing else that's blocking

production/src/workflows.ts
```diff
@@ -0,0 +1,14 @@
+// @@@SNIPSTART typescript-hello-workflow
```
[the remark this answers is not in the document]
lorensr: 😄

PR #67, opened 2021-11-24
Add instrumentation sample

bergundy:


sw-yx (approved):

instrumentation/src/worker.ts
```diff
@@ -0,0 +1,99 @@
...
+  // @@@SNIPSTART typescript-worker-full-logging-setup
+  // Create a worker that will use the Core instance installed above
+  const worker = await Worker.create({
+    workflowsPath: require.resolve('./workflows'),
+    activities,
+    taskQueue: 'instrumentation',
+    // Install interceptors
+    interceptors: {
+      activityInbound: [(ctx) => new ActivityInboundLogInterceptor(ctx, activityLogger)],
+      workflowModules: [require.resolve('./workflows/interceptors')],
```
lorensr, line 84: Should we support custom bundling?

instrumentation/src/workflows/interceptors.ts
```diff
@@ -0,0 +1,34 @@
...
+import {
+  WorkflowInterceptorsFactory,
+  WorkflowInboundCallsInterceptor,
+  WorkflowExecuteInput,
+  Next,
+} from '@temporalio/workflow';
+import { logger } from './logger';
+
+/** Logs Workflow executions and their duration */
+export class WorkflowInboundLogInterceptor implements WorkflowInboundCallsInterceptor {
```
lorensr, line 11: Does this need to be exported?

lorensr (commented):
Cool stuff! Can add to root README, and maybe add `<details>` to bottom of `instrumentation/README` with [example worker output](https://gist.github.com/lorensr/dd26d49b56b97dfe4320d1e503684d04)

instrumentation/src/workflows/interceptors.ts
bergundy: Removed the export so people don't get confused.

bergundy:
> Cool stuff! Can add to root README, and maybe add `<details>` to bottom of `instrumentation/README` with [example worker output](https://gist.github.com/lorensr/dd26d49b56b97dfe4320d1e503684d04)

Good idea, I was already going to add to the README, also added the output.

instrumentation/src/worker.ts
bergundy: Maybe we should but it'd make the sample a bit less concise.
We should perhaps consider having all-in-one dev + prod ready sample.

bergundy:
Oops accidentally pushed the otel sample fix here

PR #72, opened 2021-11-26
Fix task queue name in logging-sinks

lorensr:
## What was changed

Match `taskQueue` between execute script and worker


vkarpov15 (approved):

JoshuaKGoldberg (approved):
This does fix it for me, thanks!

PR #78, opened 2021-12-06
add dsl example

sw-yx:
## What was changed
this is a port of https://github.com/temporalio/samples-go/tree/main/dsl

## Why?
good demo, highly requested


viperfx:
Given the following input DSL

```
{
  variables: { arg1: "value1", arg2: "value2" },
  root: {
    sequence: {
      elements: [
        {
          activity: {
            name: "Activity1",
            arguments: ["arg1"],
            result: "result1",
          },
        },
      ],
    },
  },
};
```


In the Go example, this will output `Run SampleActivity1 with input [value1]` because value1 is the value of the key to the root variables.

The following code will fix the above scenario 

```
export async function DSLInterpreter(dsl: DSL): Promise<any> {
  const bindings = dsl.variables as Record<string, string>;
  return execute(dsl.root, bindings);
}

async function execute(
  statement: Statement,
  bindings: Record<string, string | undefined>
): Promise<string | undefined> {
  // const results = {} as Record<string, string | undefined>;
  // console.log(statement)
  if ("parallel" in statement) {
    await Promise.all(
      statement.parallel.branches.map((el) => execute(el, bindings))
    );
    // todo: if one activity fails we want to cancel all the others
  } else if ("sequence" in statement) {
    let result: string | undefined;
    for (const el of statement.sequence.elements) {
      result = await execute(el, bindings);
    }
    return result;
  } else {
    const activity = statement.activity;
    let args = activity.arguments || [];
    args = args.map((arg) => bindings[arg] ?? arg);
    const activityResult = await acts[activity.name](...args);
    if (activity.result) {
      bindings[activity.result] = activityResult;
    }
    return activityResult;
  }
}
```

sw-yx:
thanks for the catch! updating my PR. still will wait for @bergundy and optionally @lorensr to review

dsl-interpreter/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "watch": [
+      "src"
+    ]
+  },
+  "dependencies": {
+    "node-yaml": "^4.0.1",
+    "temporalio": "0.16.x"
+  },
+  "devDependencies": {
+    "@tsconfig/node14": "^1.0.0",
```
bergundy, line 28: We use node16 by default, where did you copy this from?

dsl-interpreter/src/activities.ts
```diff
@@ -0,0 +1,28 @@
+export async function Activity1(arg1: string): Promise<string> {
```
bergundy, line 1: Is this a new convention using pascal case for activity functions?

dsl-interpreter/src/client.ts
```diff
@@ -0,0 +1,45 @@
+import { Connection, WorkflowClient } from '@temporalio/client';
+import { DSLInterpreter, DSL } from './workflows';
+// @ts-ignore
```
bergundy, line 3: Why is this ignored?

dsl-interpreter/src/client.ts
```diff
@@ -0,0 +1,45 @@
...
+  // In production, pass options to the Connection constructor to configure TLS and other settings.
+  // This is optional but we leave this here to remind you there is a gRPC connection being established.
+
+  const client = new WorkflowClient(connection.service, {
+    // In production you will likely specify `namespace` here; it is 'default' if omitted
+  });
+
+  // Invoke the `DSLInterpreter` Workflow, only resolved when the workflow completes
+  const result = await client.execute(DSLInterpreter, {
+    args: [dslInput], // type inference works! args: [name: string]
```
bergundy, line 35: Comment is wrong

dsl-interpreter/src/script.ts
```diff
@@ -0,0 +1,11 @@
+// @ts-ignore
```
bergundy, line 1: More ignores?

dsl-interpreter/src/types.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'node-yaml';
```
bergundy, line 1: yaml doesn't have a types package?

dsl-interpreter/src/workflows.ts
```diff
@@ -0,0 +1,57 @@
+import { proxyActivities } from '@temporalio/workflow';
+// Only import the activity types
+import type * as activities from './activities';
+
+export type DSL = {
+  variables: Record<string, string>; // could be string | number?
```
bergundy, line 6: Should be `Record<string, unkown>` AFAICT since we read it from yaml

dsl-interpreter/src/workflows.ts
```diff
@@ -0,0 +1,57 @@
...
+  return execute(dsl.root, bindings);
+}
+
+async function execute(
+  statement: Statement,
+  bindings: Record<string, string | undefined>
+): Promise<string | undefined> {
+  if ('parallel' in statement) {
+    await Promise.all(statement.parallel.branches.map((el) => execute(el, bindings)));
+    // todo: if one activity fails we want to cancel all the others
```
bergundy, line 40: Why? Is this something that's done in the other SDKs' samples?

dsl-interpreter/src/workflows.ts
```diff
@@ -0,0 +1,57 @@
...
+  const bindings = dsl.variables as Record<string, string>;
+  return execute(dsl.root, bindings);
+}
+
+async function execute(
+  statement: Statement,
+  bindings: Record<string, string | undefined>
+): Promise<string | undefined> {
+  if ('parallel' in statement) {
+    await Promise.all(statement.parallel.branches.map((el) => execute(el, bindings)));
```
bergundy, line 39: Do we not return here?

dsl-interpreter/src/workflows.ts
```diff
@@ -0,0 +1,57 @@
...
+    // todo: if one activity fails we want to cancel all the others
+  } else if ('sequence' in statement) {
+    let result: string | undefined;
+    for (const el of statement.sequence.elements) {
+      result = await execute(el, bindings);
+    }
+    return result;
```
bergundy, line 46: Why return the last result?

dsl-interpreter/src/workflows.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+type Statement = { activity: ActivityInvocation } | { sequence: Sequence } | { parallel: Parallel };
+
+const acts = proxyActivities<typeof activities>({
+  startToCloseTimeout: '1 minute',
+}) as Record<string, (...args: string[]) => Promise<string | undefined>>;
+
+/** A workflow that simply calls an activity */
+export async function DSLInterpreter(dsl: DSL): Promise<any> {
```
bergundy, line 29:
```suggestion
export async function DSLInterpreter(dsl: DSL): Promise<unknown> {
```

dsl-interpreter/src/workflows.ts
```diff
@@ -0,0 +1,57 @@
...
+  const bindings = dsl.variables as Record<string, string>;
+  return execute(dsl.root, bindings);
```
bergundy, line 31: For a better stack trace:
```suggestion
  return await execute(dsl.root, bindings);
```

dsl-interpreter/package.json
sw-yx: default hello world

dsl-interpreter/src/activities.ts
sw-yx: no just felt like copying the Go example directly which names those activities.

dsl-interpreter/src/client.ts
sw-yx: node-yaml doesnt have types, and our TS setup didnt recognize the .d.ts that i put in there and i couldnt figure it out. actually would like some help here.

dsl-interpreter/src/client.ts
sw-yx:
```suggestion
    args: [dslInput],
```

dsl-interpreter/src/types.d.ts
sw-yx: not that i could find

sw-yx: i just realized i should drop down one layer to js-yaml. will do that.

dsl-interpreter/src/workflows.ts
sw-yx: yes https://github.com/temporalio/samples-go/blob/main/dsl/workflow.go#L134

dsl-interpreter/src/workflows.ts
sw-yx: i think this part of the "DSL" is underspecced - basically theres no reliance on return value for `parallel`. we store results in the bindings, thats it

dsl-interpreter/src/workflows.ts
sw-yx: to match the output of the other DSL examples

dsl-interpreter/src/workflows.ts
sw-yx: ohh thats why you do it. i always wondered why you do that since the await is optional from an execution perspective.

dsl-interpreter/src/workflows.ts
bergundy: What's the value in it with `Promise.all`? If one activity fails the whole workflow fails and the activities will be effectively forgotten, if they heartbeat after the workflow has failed they'll get cancelled automatically.

dsl-interpreter/.eslintrc.js
```diff
@@ -0,0 +1,27 @@
...
+  rules: {
+    '@typescript-eslint/no-unused-vars': [
+      'warn',
+      {
+        argsIgnorePattern: '^_',
+        varsIgnorePattern: '^_',
+      },
+    ],
+    '@typescript-eslint/no-explicit-any': 'off',
+    'object-shorthand': ['error', 'always'],
```
sw-yx, line 24:
```suggestion
    '@typescript-eslint/no-explicit-any': 'off',
    '@typescript-eslint/no-floating-promises': 'error',
    'object-shorthand': ['error', 'always'],
```

dsl-interpreter/README.md
```diff
@@ -0,0 +1,10 @@
+# DSL Interpreter Workflow
+
+This sample demonstrates how to implement a DSL workflow, based on the [Go equivalent](https://github.com/temporalio/samples-go/tree/main/dsl). In this sample, we provide 2 sample yaml files each defines a custom workflow that can be processed by this DSL workflow sample code.
+
+### Running this sample
+
+1. Make sure Temporal Server is running locally (see the [quick install guide](https://docs.temporal.io/docs/server/quick-install/)).
+1. `npm install` to install dependencies.
+1. `npm run start.watch` to start the Worker.
+1. In another terminal, `npm run workflow1` or `npm run `workflow2` to run the Workflows accordingly
```
sw-yx, line 10:
```suggestion
1. In another terminal, `npm run workflow1` or `npm run workflow2` to run the Workflows accordingly
```

Issue #79, opened 2021-12-10
[Bug] Next.JS sample state management issues


yoDon:
### Describe the bug

The Next.JS sample app has a couple react state management issues that don't have anything to do with using temporal but which do prevent the app from working as expected.

1) Purchases never show as completed - The `onClose()` function looks at the `state` variable directly instead of going through a `useRef` wrapper. This causes the contents of the `state` variable to be out of date at the time when `onClose()` is testing for an `'ORDERED'` state (it sees `'NEW'` instead of `'ORDERED'`). The implication of this issue is the `onClose()` function never sets the local state to `'CONFIRMED'` and hence the UI never shows the item as purchased. Wrapping `state` in a `useRef` and looking at the `.current` property of the ref will fix this problem.

2) Cancelling a purchase blocks any future attempted purchases of the same item - The code currently doesn't reset the `transactionId` after cancelling a purchase. After cancelling a purchase the code needs to assign a new `uuid4()` to the `transactionId`.

### To Reproduce

1. Click buy on an item
2. Wait
3. Expect to see the button text switch to `Purchased`
4. Never actually see the button text switch to `Purchased`

and

1. Click buy on an item
2. Click cancel
3. Click buy on the same item
4. See error - unable to re-purchase

PR #80, opened 2021-12-10
Next.JS sample - Fix React state management issues

yoDon:
Fixes https://github.com/temporalio/samples-typescript/issues/79

## What was changed

Wrapped `state` in a `useRef` so `onClose()` can have access to the current value of `state`.

Added a `setTransactionId` method so the `transactionId` can be updated after a purchase is cancelled.

## Why?

As described in https://github.com/temporalio/samples-typescript/issues/79

## Checklist

1. Closes https://github.com/temporalio/samples-typescript/issues/79

2. How was this tested: As described in https://github.com/temporalio/samples-typescript/issues/79

3. Any docs updates needed? No


yoDon:
Likely one for @sw-yx to CR

sw-yx (approved):
appreciated!

PR #81, opened 2021-12-11
add "no floating promises" lint rule to all ts projects

sw-yx:
## What was changed
add "no floating promises" lint rule to all ts projectsa

## Why?
best practice. too easy to forget to await things


bergundy:
Let's update all of these: https://github.com/temporalio/samples-typescript/blob/e4ecc51d2ea0e602497cf788cbab3ebc4598e58e/.scripts/copy-shared-files.mjs#L11
I think only monorepo is missing right now.

bergundy (approved):
Approved with a single comment

PR #84, opened 2021-12-31
Upgrade some examples to SDK 0.17

vkarpov15:
## What was changed

Upgraded a bunch of examples to SDK v0.17. Doesn't look like any changes are required, but I spot checked the steps in the README for each example and found one issue with signals-queries

## Why?

samples should use latest version of the SDK unless they're explicitly demonstrating using an older SDK version

## Checklist

1. Closes 

2. How was this tested:

Walked through README steps for each sample with 0.17.x

3. Any docs updates needed?


bergundy (approved):
Thanks, we should upgrade all of the samples

vkarpov15:
@bergundy I upgraded and verified the remaining samples, except for the monorepo-folders issue that I ran into #85

bergundy:
@vkarpov15 please fix the build so we can merge.

vkarpov15:
@bergundy I fixed the build in this branch and `main`. Should be good to go. Only issue is that monorepo-folders isn't upgraded yet pending #86

PR #87, opened 2022-01-11
fix pre-commit hook on ubuntu

vkarpov15:
## What was changed

`/bin/sh` -> `/bin/bash`

## Why?

See https://stackoverflow.com/questions/54055549/linux-ubuntu-set-illegal-option-o-pipefail . `git commit` always fails on my Ubuntu box.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


lorensr (approved):
Thanks!

PR #88, opened 2022-01-12
Adding devcontainer for GitHub remote-extensions and codespaces

sw-yx:
## What was changed
Adding devcontainer for GitHub remote-extensions and codespaces

image

## Why?
Run any sample in one click

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,50 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.209.5/containers/javascript-node-postgres
+// Update the VARIANT arg in docker-compose.yml to pick a Node.js version
+{
+	"name": "Samples Typescript",
+	"dockerComposeFile": "docker-compose.yml",
+	"service": "samples",
+	"workspaceFolder": "/workspace",
+	"shutdownAction": "stopCompose",
+	// Set *default* container specific settings.json values on container create.
+	"settings": {
+		"sqltools.connections": [
+			{
+				"name": "temporal",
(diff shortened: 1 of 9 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,50 @@
...
+// Update the VARIANT arg in docker-compose.yml to pick a Node.js version
+{
+	"name": "Samples Typescript",
+	"dockerComposeFile": "docker-compose.yml",
+	"service": "samples",
+	"workspaceFolder": "/workspace",
+	"shutdownAction": "stopCompose",
+	// Set *default* container specific settings.json values on container create.
+	"settings": {
+		"sqltools.connections": [
```
jbreiding, line 12: This should be removed since temporalite is being used now instead of postgres

vkarpov15 (approved):
Went through this and got it working on my machine:

image

I second the suggestion to remove postgres - my biggest concern is that setting up the remote extension was very slow, took at least 30 mins to build. The more we can remove the better :)

vkarpov15:
Also, please add `.devcontainer` to `.prettierignore`

jbreiding:
> Went through this and got it working on my machine:
> 
> image
> 
> I second the suggestion to remove postgres - my biggest concern is that setting up the remote extension was very slow, took at least 30 mins to build. The more we can remove the better :)

When you say "my machine", what are you using the get/maintain a docker context?

If you have a VM backing this, and using default docker-desktop settings, the performance would be pretty poor.

vkarpov15:
I'm not sure how to answer "what are you using to get/maintain a docker context?", I'm afraid my understanding of Docker is pretty limited. I followed the instructions here: https://code.visualstudio.com/docs/remote/containers

sw-yx:
k gonna merge it in now since i cant edit the thing otherwise

PR #98, opened 2022-02-02
Use unique workflowId for mtls sample

bergundy:
This sample is used for SDK tests against a shared cloud namespace and there's a possibility for collision.

### DIFF
--- a/hello-world-mtls/src/client.ts
+++ b/hello-world-mtls/src/client.ts
@@ -37,7 +37,11 @@ async function run({
   await connection.untilReady();
   const client = new WorkflowClient(connection.service, { namespace });
   // Run example workflow and await its completion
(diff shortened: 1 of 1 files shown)


hello-world-mtls/src/client.ts
```diff
@@ -37,7 +37,11 @@ async function run({
   await connection.untilReady();
   const client = new WorkflowClient(connection.service, { namespace });
   // Run example workflow and await its completion
-  const result = await client.execute(example, { taskQueue, workflowId: 'my-business-id', args: ['Temporal'] });
+  const result = await client.execute(example, {
+    taskQueue,
+    workflowId: `my-business-id-${Date.now()}`,
```
lorensr, line 42: uuid would be less chance of collision, but I judge millisecond as unlikely to be an issue 😄

lorensr (approved):

PR #99, opened 2022-02-05
Add new replay sample

sw-yx:
## What was changed

as above

## Why?

teach people how to use the new replayer functionality 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


sw-yx:
want to add docs and video before merging

replay-history/src/workflows.ts
```diff
@@ -0,0 +1,31 @@
...
+  add: wf.defineSignal<[number]>('add'),
+  sub: wf.defineSignal<[number]>('sub'),
+  div: wf.defineSignal<[number]>('div'),
+  mul: wf.defineSignal<[number]>('mul'),
+  inverseFraction: wf.defineSignal<[]>('inverseFraction'),
+};
+
+export async function calculator(): Promise<void> {
+  let value = 0;
+  wf.setHandler(signals.add, (x: number) => {
```
Sushisource, line 14: I'd strongly suggest using a workflow that emits at least one command - users will want to experiment with replay failure, but this workflow is actually very very hard to make fail replay because added commands will not conflict with anything and the workflow never terminates. See https://www.notion.so/temporalio/A-fascinating-digression-on-workflow-compatibility-and-determinism-82c0582bf06e4152a6b08dc8d8f67463

sw-yx: Roey's goal for this demo evolved towards just demonstrating step through debugging rather than demonstrating failure and then fixing it, but ofc i welcome suggestions

replay-history/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "execMap": {
+      "ts": "ts-node"
+    },
+    "ext": "ts",
+    "watch": [
+      "src"
+    ]
+  },
+  "dependencies": {
+    "temporalio": "0.17.x"
```
sw-yx, line 24:
```suggestion
    "temporalio": "0.18.x"
```

Issue #101, opened 2022-02-10
[Feature Request] Send heartbeat details to next activity invocation


lorensr:
As described here: https://typescript.temporal.io/api/classes/activity.context/#heartbeat

Add to this sample: https://github.com/temporalio/samples-typescript/blob/main/activities-cancellation-heartbeating/src/workflows.ts

Thanks @mabadir for reporting

lorensr:
From the docs, I imaginge it would be something like:

```ts
export async function runCancellableActivity(): Promise<void> {
  try {
    await fakeProgress();
  } catch (err) {
    if (err instanceof ActivityFailure && err.cause instanceof TimeoutFailure) {
      await fakeProgress(err.cause.lastHeartbeatDetails)
    } else {
      throw err;
    }
  }
}
```

https://typescript.temporal.io/api/classes/client.timeoutfailure/#lastheartbeatdetails

sw-yx:
i think the design pattern we need to work out is how to loop this activity until completion, handling multiple failures until an expected timeout. 

as a step 2, perhaps this would then become a first class Activity API where the SDK would handle it for the user instead of having to write all this heartbeat boilerplate

Sushisource:
I think the much easier way to solve this problem is to set an appropriate retry policy. If you let the activity fail, and then say "oh but I actually wanted to pass the progress on and invoke it again", well, it seems like your retry policy wasn't really what you meant. What you actually wanted here was for it to just keep retrying.

mjameswh:
Fixed in #102

PR #102, opened 2022-02-11
improve guidance for heartbeating

sw-yx:
## What was changed
improve guidance for heartbeating

## Why?
solve #101


lorensr:
I added instructions for exercising your code. Although anyone know why `ActivityTaskStarted`, which happened at 20s, appears later in the history than `WorkflowExecutionCancelRequested`, which happens at 40s?

image

sw-yx:
not sure, we can get on a call with Roey about it

PR #107, opened 2022-03-11
Improve .post-create message

lorensr:
Based on this user support question: https://temporalio.slack.com/archives/CTRCR8RBP/p1647015491342459

If we like this, I'll propagate to sample directories before merging.

### DIFF
--- a/.shared/.post-create
+++ b/.shared/.post-create
@@ -1,10 +1,18 @@
-To begin development, start Temporal Server:
+To begin development, install Docker:
 
-  {cyan cd} ~/path/to/temporal/docker-compose/
+  {cyan brew cask install docker}
+  {cyan open} /Applications/Docker.app
+  {cyan brew install docker-compose}
+
+And start Temporal Server:
+
+  {cyan git clone} https://github.com/temporalio/docker-compose.git ~/temporal-server
+  {cyan cd} ~/temporal-server
   {cyan docker-compose up}
 
-{dim.italic If you haven’t run Temporal Server before, visit:
-https://docs.temporal.io/docs/typescript/getting-started/}
+Make sure you're using Node version 14+:
+
+  {cyan brew install node@16}
 
 Then, in the project directory, using two other shells, run these commands:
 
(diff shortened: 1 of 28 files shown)


bergundy (commented):
Why is this Mac specific?

lorensr:
Idea is to make it simplest for those that need clearest instructions. My sense from JS ecosystem is that most devs, regardless of their OS, would be used to seeing `brew install` instructions, and know to do the equivalent for their OS. 

I could certainly change to links to download pages tho. @sw-yx thoughts?

cretz:
IMO we don't need to tell people how to install docker, and especially shouldn't put OS-specific instructions without acknowledging it's for a specific OS. Even the `cd ~/temporal-server` is platform specific.

vkarpov15:
I don't like the idea of putting OSX-specific instructions, because it implies that this only works on OSX.

nelsonic:
Just chiming as a recent adopter of Temporal and trying to on-board a team of engineers ... 
Agree that having MacOS-specific instructions could give the impression that it's MacOS-only. 
However _not_ having _any_ instructions means some people get stuck and have to Google
i.e. increase friction / barriers to adoption.
It might be easier to just add a line: 
This works on all platforms. If you're not using MacOS, Help us by extending these docs with instructions for your OS.

bergundy:
I'm fine with *adding* MacOS specific instructions and possibly customizing the post create message based on the OS (although the latter a bit overkill and I wouldn't say it is high priority now).

.shared/.post-create
```diff
@@ -1,10 +1,18 @@
-To begin development, start Temporal Server:
+To begin development, install Docker:
 
-  {cyan cd} ~/path/to/temporal/docker-compose/
+  {cyan brew cask install docker}
+  {cyan open} /Applications/Docker.app
+  {cyan brew install docker-compose}
```
sw-yx, line 5: i have a dumb question. how expensive would it be to check if docker is already installed, and/or if docker daemon is already running? feel like i'd like to get rid of this msg if its unnecessary.

sw-yx (approved):
left a comment but lgtm otherwise

.shared/.post-create
lorensr: Don't know but good idea! Added:

https://github.com/temporalio/sdk-typescript/issues/541

PR #109, opened 2022-03-18
Fix duplicate CI

lorensr:


lorensr:
Now does this:

image

instead of:

image

PR #112, opened 2022-03-24
Add custom payload converter sample

lorensr:


README.md
```diff
@@ -71,6 +71,7 @@ Each directory in this repo is a sample Temporal project built with the [TypeScr
 - [**Patching**](https://docs.temporal.io/docs/typescript/patching/): Patch in new Workflow code when making updates to Workflows that have executions in progress in production.
 - [**Logging**](https://github.com/temporalio/samples-typescript/tree/main/logging-sinks): Use Sinks to extract data out of Workflows for logging/metrics/tracing purposes.
 - [**Instrumentation**](https://github.com/temporalio/samples-typescript/tree/main/instrumentation): Use a [winston](https://github.com/winstonjs/winston) logger to get logs out of all SDK components and get metrics and traces out of Rust Core.
+- [**Custom Payload Converter**](https://github.com/temporalio/samples-typescript/tree/main/ejson): Customize data serialization by creating a `PayloadConverter`.
```
bergundy, line 74: Maybe mention something about EJSON and supported types (like Date)

ejson/src/client.ts
```diff
@@ -0,0 +1,39 @@
...
+    dataConverter: { payloadConverterPath: require.resolve('./payload-converter') },
+  });
+  // @@@SNIPEND
+
+  // @@@SNIPSTART typescript-ejson-client
+  const user: User = {
+    id: uuid(),
+    // age: 1000n, BigInt isn't supported
+    hp: Infinity,
+    matcher: /.*Stormblessed/,
```
bergundy, line 18: Have you read them all??

ejson/src/payload-converter.ts
```diff
@@ -0,0 +1,5 @@
+// @@@SNIPSTART typescript-ejson-converter
+import { EjsonPayloadConverter } from './ejson-payload-converter';
+
+export const payloadConverter = new EjsonPayloadConverter();
```
bergundy, line 4: We might want to show how to use the composite converter here instead of just a single one.

ejson/src/ejson-payload-converter.ts
```diff
@@ -0,0 +1,54 @@
...
+export class EjsonPayloadConverter implements PayloadConverterWithEncoding {
+  public encodingType = 'ejson/plain' as EncodingType;
+
+  public toPayload(value: unknown): Payload | undefined {
+    if (value === undefined) return undefined;
+    let ejson;
+    try {
+      ejson = EJSON.stringify(value);
+    } catch (e) {
+      throw new UnsupportedEjsonTypeError(
```
bergundy, line 26: You should have returned `undefined` here and used the composite converter which throws.

bergundy (commented):
Sorry for the late review

ejson/src/client.ts
lorensr: Everything in the Cosmere but the graphic novels!

ejson/src/payload-converter.ts
lorensr: Ah, it doesn't support `undefined`. Will add

ejson/src/ejson-payload-converter.ts
lorensr: I think it makes sense as a descriptive error for the last converter in a composite—it's what we do in Json

ejson/src/client.ts
bergundy: I only read the first 3 in the Stormlight Archive

ejson/src/ejson-payload-converter.ts
bergundy: I thought we dropped that for JSON..
It makes it so you can't reorder the converters.

ejson/src/client.ts
lorensr: Number 4 is out, and I also really liked 2.5 and 3.5 (Edgedancer and Dawnshard)

ejson/src/ejson-payload-converter.ts
lorensr: I thought I had convinced you, but don't fully remember. I think having a good error message is more impt than ordering flexibility, but maybe we can put the error in Composite.

Right now the `Composite.toPayload @throws` is lying—it always returns a value or undefined. I don't remember if there was as reason for that, but if not, can change to throw descriptive error.

image

lorensr: If it threw, I think we'd just have to catch and use `undefined` in a lot of places—we could use a helper

lorensr: `@throws` isn't lying, sorry thought it was `UnsupportedTypeError`

bergundy: The composite converter should be the one that throws not JSON.
This was part of the review.
We should fix this issue before the next release.

lorensr: Fixed in https://github.com/temporalio/sdk-typescript/pull/558

activities-examples/package.json
```diff
@@ -24,13 +24,15 @@
   "dependencies": {
     "axios": "0.21.1",
     "node-fetch": "2.x",
-    "temporalio": "0.19.x"
+    "temporalio": "0.19.x",
+    "uuid": "^8.3.2"
```
joebowbeer, line 28: @lorensr move `uuid` to devDependencies?

lorensr: Ah thanks! https://github.com/temporalio/samples-typescript/pull/157

PR #118, opened 2022-04-03
Add encryption sample

lorensr:
Uses `@ronomon/crypto-async`, which unlike `crypto` does work in a thread pool. Used the same algorithm as Java and Go encryption samples, so it should be compatible, but I plan on adding tests to `sdk-features`.

`ts-ignore`ing due to bug in types, pending merge of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59580


encryption/.post-create
```diff
@@ -0,0 +1,20 @@
+To begin development, install Docker:
```
cretz, line 1: Do we have to have all of these files for every sample? If we want to change the content, will we be updating them all? Same goes for the dependencies in package.json, are we going to have to keep dozens of dependency files up to date? Neither Java nor Go samples have to.

encryption/README.md
```diff
@@ -0,0 +1,26 @@
+# Encryption
+
+Create a custom data converter that encrypts data with AES. See [encryption docs](https://docs.temporal.io/docs/typescript/data-converters#encryption).
+
+## Code
+
+- `EncryptionCodec`: [encryption-codec.ts](https://github.com/temporalio/samples-typescript/blob/main/encryption/src/encryption-codec.ts)
```
cretz, line 7: Can use relative paths here (granted unsure about npm, but I figure we aren't putting each sample there as its own project anyways)

cretz (approved):

encryption/.post-create
lorensr: We have a lot of config files like this that are duplicated in git. Some have modifications, in which case they're excluded from the pre-commit copy script: 

https://github.com/temporalio/samples-typescript/blob/main/.scripts/copy-shared-files.mjs

Separate `package.json`s so they work standalone, eg via our command-line tool, `npx @temporalio/create@latest --sample hello-world`

lorensr: Better design would be for each directory have a `.copy-config` file, rather than listing them in the script

PR #131, opened 2022-04-12
use nanoid for workflow ids

### LINKED ISSUE #110: [ Documentation request ]  Better workflowId generator
# Brief description

I suggest _**not**_ using `random * 1000` to generate workflow IDs in the documentation and sample code. For example:

```JavaScript
workflowId: 'wf-id-' + Math.floor(Math.random() * 1000),
```

This appears in two samples and a blog:

* https://github.com/temporalio/samples-typescript/blob/main/hello-world/src/client.ts#L21
* https://github.com/temporalio/samples-typescript/blob/main/grpc-calls/src/client.ts#L5
* https://github.com/temporalio/documentation/blob/master/blog/2022-02-28-temporal-rest.md?plain=1#L57

In only one instance is there a disclaimer comment:

```JavaScript
// in practice, use a meaningful business id, eg customerId or transactionId
```

Note that these IDs are required to be unique among open workflow executions in the same namespace.

Calculating the odds of a collision is solving the generalized birthday problem. With 1000 IDs to choose from, there's a 50% chance of a collision after generating 38 IDs, and a greater than 99% chance of collision after generating 100 IDs. (Assuming `random` is random.)

#  Your recommended content

1. I think the disclaimer should always be present.
1. If an ID generator is used in these samples, I suggest using something better than `random * 1000`, as I expect this snippet will end up being copied further. Use NanoID, UUID, or similar? Or just `'my-business-id'`?

vkarpov15:
Fix #110

## What was changed

Replaced `Math.random()` with [nanoid](https://www.npmjs.com/package/nanoid) for randomized Workflow ids.

## Why?

I prefer to use randomized workflow ids in samples because I like to be able to run a sample multiple times when figuring out what it does. If a sample has a hard-coded workflow id, that sample will throw an error if you try to run it multiple times, unless you change the id yourself.

As @joebowbeer pointed out, using `Math.random()` to generate workflow ids is bad practice. The risk of collisions is too high for a production application. And our samples shouldn't imply otherwise.

`nanoid` is a good alternative for generating concise, randomized ids. I'm thinking it is better than uuid because the resulting id is considerably shorter, and so leads to less ugly workflow ids in the Temporal UI.

However, I'm also considering using the built-in [crypto.randomUUID() function](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID). Using a built-in function would save us an extra dependency, but `randomUUID()` has longer output and is only supported in Node >= 16.7.0, which is slightly off from our currently supported Node version range.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/grpc-calls/package.json
+++ b/grpc-calls/package.json
@@ -20,6 +20,7 @@
     ]
   },
   "dependencies": {
+    "nanoid": "3.x",
     "temporalio": "0.20.x"
   },
   "devDependencies": {
--- a/grpc-calls/src/client.ts
+++ b/grpc-calls/src/client.ts
@@ -1,9 +1,10 @@
 import { Connection } from '@temporalio/client';
 import { defaultPayloadConverter, toPayloads } from '@temporalio/common';
+import { nanoid } from 'nanoid';
 
 async function run() {
-  const workflowId = 'wf-id-' + Math.floor(Math.random() * 1000);
-  const requestId = 'request-id-' + Math.floor(Math.random() * 1000);
+  const workflowId = 'workflow-' + nanoid();
+  const requestId = 'request-' + nanoid();
   // @@@SNIPSTART typescript-grpc-call-basic
   const connection = new Connection();
 
--- a/hello-world/package.json
+++ b/hello-world/package.json
@@ -20,6 +20,7 @@
     ]
   },
   "dependencies": {
+    "nanoid": "3.x",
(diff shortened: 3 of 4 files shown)


joebowbeer (approved):
LGTM

lorensr:
👍 shorter than uuid

image

PR #132, opened 2022-04-13
Add to production sample

lorensr:
- Add Dockerfile and shared connectionOptions
- Add HTTP server to provide status
- Don't copy certs to image on build

Adding instrumentation and more README updates in next PR

### DIFF
--- a/production/Dockerfile
+++ b/production/Dockerfile
@@ -1,0 +1,20 @@
+# syntax=docker/dockerfile:1
+
+FROM node:16.14.2-bullseye
+
+RUN apt update && apt install -y ca-certificates
+
+ENV NODE_ENV=production
+WORKDIR /app
+
+COPY ["package.json", "./"]
+RUN npm install --production
+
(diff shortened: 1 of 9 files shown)


vkarpov15 (approved):

production/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt update && apt install -y ca-certificates
+
+ENV NODE_ENV=production
+WORKDIR /app
+
+COPY ["package.json", "./"]
+RUN npm install --production
+
+ARG TEMPORAL_SERVER="host.docker.internal:7233"
```
bergundy, line 13: Why put these in args?
It's typically something you'd provide at container run time, not bake into the image since this image could be used in different environments.

bergundy (changes requested):
Not loving this sample.
It's not a complete production sample because it omits logging / metrics setup and has a lot of duplication with the mtls sample.

PR #133, opened 2022-04-14
add no-restricted-imports rule for workflows in samples to disallow importing node built-ins

### LINKED ISSUE #106: [Feature Request] Lint against Node imports only inside Workflow files

we should be able to use this: https://eslint.org/docs/rules/no-restricted-imports

however it should not be applied indiscriminately, only on workflow files

ideally to keep it lightweight, we just add a comment at the top of a workflow file to keep us honest.

### Additional context

https://twitter.com/ryanflorence/status/1499456892724137987?s=20&t=2MMXl5duaDr4OiIw2U2W9A

vkarpov15:
Fix #106

## What was changed

Add [ESLint no-restricted-imports](https://eslint.org/docs/rules/no-restricted-imports) rule on all the modules listed here: https://github.com/temporalio/sdk-typescript/blob/0310ad6e44ea49220e3ac14fa90c3e8c78ec3647/packages/worker/src/workflow/bundler.ts#L11

## Why?

Importing these modules can lead to hard-to-grok runtime errors, like:

```
  2022-04-14T18:02:11.115953Z  WARN temporal_sdk_core::worker: Failing workflow activation, run_id: "9ba88cd5-777c-44ee-ab9c-2aef323e0901", failure: Failure { failure: Some(Failure { message: "util_1.default.inspect is not a function", source: "TypeScriptSDK", stack_trace: "TypeError: util_1.default.inspect is not a function\n    at example (webpack-internal:///./src/workflows.ts:15:39)", cause: None, failure_info: None }) }
```

I think it is worthwhile to disallow these in the default lint config. 

Ideally, I'd like to export [this list](https://github.com/temporalio/sdk-typescript/blob/0310ad6e44ea49220e3ac14fa90c3e8c78ec3647/packages/worker/src/workflow/bundler.ts#L11) from the TypeScript SDK and use that. What do you think @lorensr?

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


lorensr (approved):
Looks good! Yes, I'll export from SDK

image

cretz:
Can we not share any of these files/configs? Must every sample have over a half dozen copy/pasted metadata files?

vkarpov15:
@cretz we do share these files as best we can: there's a copy-shared-files script that copies this file from `.scripts` to every sample. The reason why we structured the samples repo this way is to keep each sample isolated from the others, so developers can easily copy a full sample into a separate git repo and start working without untangling cross-sample dependencies.

cretz:
@vkarpov15 - makes sense. In the samples for every other SDK (Go, Java, PHP, Python), we accept that samples are not standalone, they are meant to be viewed as sample code, and we treat it as one large project accepting that dependencies are shared across them even if only one may use the dependency. We haven't really had a requirement that users be expected to copy a full sample into a separate git repo. But it's no big deal if we keep that requirement for this SDK.

lorensr:
The TS project initializer CLI also copies the subdirectory into a new git
repo

lorensr:
Let's use this method: 

https://github.com/temporalio/sdk-typescript/pull/591#issuecomment-1101930061

and also disallow the prefixed versions, like `node:fs`.

vkarpov15:
@lorensr take a look at my latest change. I got rid of the hard-coded list in favor of `builtinModules`, and added a check for the `node:` prefix. Also added a hardcoded list of ok built-in modules, currently just `assert`

lorensr (approved):

PR #137, opened 2022-04-28
Add codec server to encryption sample

bergundy:


encryption/src/codec-server.ts
```diff
@@ -0,0 +1,91 @@
...
+  metadata?: Record<string, string> | null;
+  data?: string | null;
+}
+
+interface Body {
+  payloads: JSONPayload[];
+}
+
+/**
+ * Helper function to converta valid proto JSON to a payload object.
```
robholland, line 18:
```suggestion
 * Helper function to convert a valid proto JSON to a payload object.
```

encryption/src/codec-server.ts
```diff
@@ -0,0 +1,91 @@
...
+async function main(port = 8888) {
+  const converter = await EncryptionCodec.create('test-key-id');
+
+  const app = express();
+  app.use(cors({ origin: 'http://localhost:8088', allowedHeaders: ['x-namespace', 'content-type'] }));
+  app.use(express.json());
+
+  app.post('/decode', async (req, res) => {
+    try {
+      const { payloads: raw } = req.body as Body;
```
robholland, line 56: I think it may be safer to use `IPayloads` here rather than rely implicitly on it's serialisation.

encryption/src/codec-server.ts
```diff
@@ -0,0 +1,91 @@
...
+      metadata &&
+      Object.fromEntries(
+        Object.entries(metadata).map(([k, v]): [string, string] => [k, Buffer.from(v).toString('base64')])
+      ),
+    data: data ? Buffer.from(data).toString('base64') : undefined,
+  };
+}
+
+async function main(port = 8888) {
+  const converter = await EncryptionCodec.create('test-key-id');
```
robholland, line 48:
```suggestion
  const codec = await EncryptionCodec.create('test-key-id');
```

robholland: I think it's best to try and avoid confusion between codec and converter.

bergundy: Agree, I missed this, probably copied from some other file.

encryption/src/codec-server.ts
bergundy: Not sure what you mean?
This version of `@temporalio/proto` doesn't support proto JSON serialization so I had to hack it.
We will support it in the next release though.

robholland: I mean that the API is defined in terms of Payloads, not Payload.

robholland (approved):
Looks good to me.

encryption/src/codec-server.ts
bergundy: Sure, I guess i just named it `Body` here and technically we only care about the `Payload[]`, that's what the codec handles.

bergundy:
Thanks @robholland

bergundy:
@robholland we should add oauth integration here too, want to take this on or guide me or @lorensr?

robholland:
I can take a stab at it.

robholland:
Fixes #137

PR #141, opened 2022-05-07
Use json/plain for EJSON sample

lorensr:


ejson/src/ejson-payload-converter.ts
```diff
@@ -33,7 +34,8 @@ export class EjsonPayloadConverter implements PayloadConverterWithEncoding {
 
     return {
       metadata: {
-        [METADATA_ENCODING_KEY]: u8('ejson/plain'),
+        [METADATA_ENCODING_KEY]: u8('json/plain'),
+        format: u8('extended'),
```
lorensr, line 38: In case user wants to create their own Composite converter that can differentiate between EJSON and JSON

bergundy: Let’s explain this in code and add an explanation in the readme of how to migrate from default converter

vkarpov15 (approved):

bergundy (approved):

PR #142, opened 2022-05-10
Update comment in worker.ts

guychouk:
Renamed the queue from `tutorial` to `hello-world` in the comment


lorensr:
Thanks for the fix! 😊

PR #145, opened 2022-05-17
Saga pattern sample

mnichols:
## What was changed
A very simple `saga` implementation was provided.

## Why?
This helps show one way Temporal can be used to implement this common orchestration pattern.

## Checklist

2. How was this tested: Manually

3. Any docs updates needed? README shows how to use


saga/src/worker.ts
```diff
@@ -0,0 +1,81 @@
...
+  const worker = await Worker.create(opts);
+
+  await worker.run();
+}
+
+run(getEnv()).catch((err) => {
+  console.error(err);
+  process.exit(1);
+});
+// @@@SNIPEND
```
lorensr, line 63:
```suggestion
```

mnichols: removed!

lorensr:
`npm run build` is failing for me:
image

saga/README.md
```diff
@@ -0,0 +1,76 @@
+# Saga
+
+This sample demonstrates orchestrating microservices using a very simplistic Saga pattern.
+
+## Use Case
+
+The act of `OpenAccount` for a company shows coordinating between Clients, Accounts, Banking, and
```
lorensr, line 7:
```suggestion
The `openAccount` Workflow opens a new bank account. It coordinates between the Clients, Accounts, Banking, and
```

hello-world/src/activities.ts
```diff
@@ -1,5 +1,6 @@
 // @@@SNIPSTART typescript-hello-activity
 export async function greet(name: string): Promise<string> {
+  console.log('ACTIVITY LOG', name);
```
bergundy, line 3: This goes as a snippet into our docs, let's remove the console.log

saga/src/activities.ts
```diff
@@ -0,0 +1,17 @@
+import { BoundedContextClients } from './types/clients';
+
+// createActivities just returns client implementations directly right now, but later might decorate
+export function createActivities(clients: BoundedContextClients) {
+  return {
+    createAccount: clients.accounts.createAccount,
```
bergundy, line 6: Shouldn't these methods be bound to the instance?

saga/src/activities.ts
```diff
@@ -0,0 +1,17 @@
+import { BoundedContextClients } from './types/clients';
+
+// createActivities just returns client implementations directly right now, but later might decorate
+export function createActivities(clients: BoundedContextClients) {
```
bergundy, line 4: This is completely optional, you could just use call `proxyActivities` for each individual client class. Up to you.

saga/src/clients/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    return Promise.resolve(undefined);
+  }
+}
+class PostOfficeClient implements PostOffice.Client {
+  addAddress(params: Commands.AddPostalAddress): Promise<void> {
+    if (params.shouldThrow) {
+      throw new Error(params.shouldThrow);
+    }
+    console.log('ADDING ADDRESS', params);
+    return Promise.resolve(undefined);
```
bergundy, line 47: Instead of the resolve calls, you can make the method `async addAddress...`

saga/src/env.ts
```diff
@@ -0,0 +1,48 @@
+// Helpers for configuring the mTLS client and worker samples
```
bergundy, line 1: @lorensr we should probably reuse this utility in other samples, it would be useful to show users how to set all of this up.

hello-world/src/activities.ts
mnichols: oof this shouldnt have been in there at all. removing

saga/src/activities.ts
mnichols: i mean, they can be, but in this case it really is just illustrating a common pattern of delegating to something like a HTTP client as an "activity" and compose all the clients funcs for use in workflows. you want me to change it to something in particular? Happy to do so

saga/src/types/clients.d.ts
```diff
@@ -0,0 +1,42 @@
+import { Commands } from './commands';
```
bergundy, line 1: Curious why this is required, just better organization?
Also, please make this `.ts` instead of `.d.ts`.

saga/src/types/commands.d.ts
```diff
@@ -0,0 +1,45 @@
+import { BankDetails, PostalAddress } from './values';
```
bergundy, line 1: Shouldn't be `.d.ts`

saga/src/worker.ts
```diff
@@ -0,0 +1,81 @@
...
+          console.error('workflow: ', workflowInfo.runId, 'message: ', message);
+        },
+        callDuringReplay: false, // The default
+      },
+    },
+  };
+
+  // registrations
+  const singletonClients = await createClients();
+  const activities = createActivities(singletonClients) as any;
```
bergundy, line 29: Curious why you had to cast to any

bergundy: It'd be better to avoid that

saga/src/worker.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const opts: WorkerOptions = {
+    workflowsPath: require.resolve('./workflows'),
+    activities,
+    taskQueue: env.taskQueue,
+    sinks,
+  };
+
+  if (isRemoteEnv(env)) {
```
bergundy, line 38: This sample is already pretty big, I'd consider removing the complex setup from this.

saga/src/env.ts
bergundy: Thinking about this some more, I'd rather put this "advanced" setup in a dedicated sample to simplify.

saga/src/worker.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  await worker.run();
+}
+
+run(getEnv()).catch((err) => {
+  console.error(err);
+  process.exit(1);
+});
+
+export default async function createWorkerClient(env: Env) {
```
bergundy, line 65: Why export this?

saga/src/types/clients.d.ts
mnichols: this is explicitly defining messaging contracts...where i have been before we defined `.d.ts` files for such API contracts. 
is that what you are asking? I may misunderstand..happy to do something different if you prefer.

saga/src/workflows.ts
```diff
@@ -0,0 +1,119 @@
...
+    // successfully called, so clear if a failure occurs later
+    compensations.unshift({
+      message: prettyErrorMessage('reversing add bank account'),
+      fn: () => disconnectBankAccounts({ accountId: params.accountId }),
+    });
+  } catch (err) {
+    if (err instanceof TemporalFailure && err.cause instanceof ApplicationFailure) {
+      logger.err(err.cause.message);
+    } else {
+      logger.err('error while opening account');
```
bergundy, line 89: Probably don't want to hide the original message

saga/src/workflows.ts
```diff
@@ -0,0 +1,119 @@
...
+      details: params.bankDetails,
+      shouldThrow: prettyErrorMessage('add bank account failed'),
+    });
+    // successfully called, so clear if a failure occurs later
+    compensations.unshift({
+      message: prettyErrorMessage('reversing add bank account'),
+      fn: () => disconnectBankAccounts({ accountId: params.accountId }),
+    });
+  } catch (err) {
+    if (err instanceof TemporalFailure && err.cause instanceof ApplicationFailure) {
```
bergundy, line 86:
```suggestion
    if (err instanceof ActivityFailure && err.cause instanceof ApplicationFailure) {
```

saga/src/worker.ts
mnichols: I think `proxyActivities` was choking on the explicit type being returned

saga/src/workflows.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+}
+
+function prettyErrorMessage(message: string, err?: any) {
+  let errMessage = '';
+  if (err && err instanceof TemporalFailure) {
+    errMessage = `${err.cause?.message}`;
+  }
+  if (err && err instanceof Error) errMessage = err.message;
```
bergundy, line 117: This overwrites the errMessage set previously.
I'd also replace TemporalFailure with ActivityFailure.

saga/src/worker.ts
mnichols: initially I was trying to unify connection creation...it isn't clear to me why they are separate for a worker versus client, but then I had to change tactic...it is unnecessary and will remove

saga/src/env.ts
mnichols: by `advanced` you mean the toggle between remote and local connectivity?

saga/src/activities.ts
mnichols: Yes you are right...when I was demoing this I was going over how activities evolve though and end up oftimes decorating something like the underlying HTTP or gRPC client (because of param serialization for example). I intentionally kept some of this ceremonial to illustrate even if it is extraneous. 

Do you want me to lean it up in this repo though?

mnichols:
@bergundy @lorensr I think I've implemented the fixes you requested (thank you!) . I have some questions for the outstanding ones and am happy to lean it up more if you would like.
1. only support local connectivity here ? 
2. use client implementations directly instead of composing them

saga/src/activities.ts
bergundy: The issue with the way you have it set up is that if someone extends this sample and tries to reference `this` in any of the client methods, they'll get the global object instead of the instance because the methods aren't bound.

You can bind them with:
```ts
createAccount: clients.accounts.createAccount.bind(clients.accounts)
-- OR --
createAccount: (...args: Parameters<typeof clients.accounts.createAccount>) => clients.accounts.createAccount(...args)
```

The former is obviously cleaner :)

saga/src/worker.ts
bergundy: This should be related to `proxyActivities`, please remove and report if something isn't working as expected.

bergundy (approved):
Approved with a few comments.
TBH I don't understand why we put so much effort into abstracting `try ... catch` everywhere but it's nice to have this sample in TS to align with other langs.

saga/src/activities.ts
mnichols: implemented

saga/src/env.ts
mnichols: i definitely agree..removed and isolated in a separate tool

saga/src/worker.ts
mnichols: Previously I had returned a interface type from `createActivities` but that was removed. I have dropped it here now.

PR #150, opened 2022-05-25
Update search attribute sample

lorensr:
Doesn't build yet. I'll test and merge after this is merged & released: https://github.com/temporalio/sdk-typescript/pull/657


search-attributes/src/workflows.ts
```diff
@@ -1,3 +1,14 @@
+import { upsertSearchAttributes, workflowInfo } from '@temporalio/workflow';
+
+// @@@SNIPSTART typescript-search-attributes-workflow
 export async function example(): Promise<string> {
-  return 'hello workflow';
+  const customInt = (workflowInfo().searchAttributes?.CustomIntField[0] as number) || 0;
+  upsertSearchAttributes({
```
bergundy, line 6: Show how to delete (set to `[]`) and show that you can read modified search attributes

search-attributes/src/workflows.ts
```diff
@@ -1,3 +1,14 @@
+import { upsertSearchAttributes, workflowInfo } from '@temporalio/workflow';
+
+// @@@SNIPSTART typescript-search-attributes-workflow
 export async function example(): Promise<string> {
-  return 'hello workflow';
+  const customInt = (workflowInfo().searchAttributes?.CustomIntField[0] as number) || 0;
```
bergundy, line 5: Aren't search attributes always defined on workflowInfo? They should be

search-attributes/src/client.ts
```diff
@@ -4,24 +4,25 @@ import { example } from './workflows';
...
       CustomBoolField: [true],
-      CustomDoubleField: [3.14],
       CustomDatetimeField: [new Date()],
       CustomStringField: [
         'String field is for text. When query, it will be tokenized for partial match. StringTypeField cannot be used in Order By',
       ],
     },
   });
+
+  const { searchAttributes } = await handle.describe();
```
bergundy, line 22: Wait for workflow to complete first

search-attributes/src/workflows.ts
lorensr: Shall we set to empty object on start? 
image

search-attributes/src/client.ts
lorensr: This snippet is just meant to show start & describe—not that upsert works.

lorensr: Added more output below this line to clarify

bergundy: We typically don't leave workflows hanging in samples and tests

search-attributes/src/workflows.ts
bergundy: Yes

search-attributes/src/client.ts
bergundy: I see you're now waiting for workflow to complete

search-attributes/src/client.ts
```diff
@@ -4,24 +4,26 @@ import { example } from './workflows';
...
     searchAttributes: {
-      CustomIntField: [2], // update CustomIntField from 1 to 2, then insert other fields
-      CustomKeywordField: ['Update1'],
+      CustomIntField: [2],
+      CustomKeywordField: ['keywordA', 'keywordB'],
       CustomBoolField: [true],
-      CustomDoubleField: [3.14],
       CustomDatetimeField: [new Date()],
       CustomStringField: [
         'String field is for text. When query, it will be tokenized for partial match. StringTypeField cannot be used in Order By',
```
bergundy, line 17:
```suggestion
        'String field is for text. When queried, it will be tokenized for partial match. StringTypeField cannot be used in Order By',
```

bergundy (approved):
We should also show how to create search attributes using the operator API

search-attributes/src/workflows.ts
bergundy: Looking at the sample code, I'd make the SearchAttributes type `Record<string, SearchAttributeValue | undefined>` so TS doesn't assume that any key will be present in the map

lorensr: Yeah. https://github.com/temporalio/sdk-typescript/pull/738

PR #160, opened 2022-06-25
Use Connection.connect in hello-world-js

lorensr:


bergundy (approved):
Thanks!
Wanna fix the lint warnings while you’re at it?

vkarpov15 (approved):
Is there any more info on when you should use `connect()` vs `create()`? And should we make this change in other samples?

lorensr:
Only non-deprecated methods left in API are `Connection.connect` and `NativeConnection.connect`. All remaining .creates are changed in this PR.

PR #164, opened 2022-07-01
Install ts-node in hello-world sample

bergundy:


rylandg (approved):
LGTM, thanks for quick fix

PR #169, opened 2022-07-12
Add NestJS exchange rates example

vkarpov15:
## What was changed

Porting https://github.com/vkarpov15/typescript-starter-with-temporal into the samples repo. Most of the code is from [Nest's TypeScript starter](https://github.com/nestjs/typescript-starter), but I also added a Temporal Workflow.

@joebowbeer I'd appreciate your comments here too.

## Why?

[NestJS is very popular](https://www.npmjs.com/package/@nestjs/core). It would be helpful to at least have a sample app, and perhaps move in the direction of having an officially supported Nest wrapper for Temporal analogous to [`@nestjs/mongoose`](https://www.npmjs.com/package/@nestjs/mongoose)

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy:
Did you consider using https://github.com/KurtzL/nestjs-temporal?

Aren’t the examples in that readme enough for showing the integration between the 2 frameworks?

vkarpov15:
@bergundy I hadn't seen that repo before I started working on this example. Whether we use nestjs-temporal or not, I do think it would be helpful to have a NestJS sample. Nest is a fairly opinionated framework, and in my experience Nest developers tend to look for Nest-specific samples and libs more so than devs who use other frameworks

bergundy:
I'm really on the fence here, we have so many samples already.
It's becoming a maintenance burden.

I guess to your point there's already a project that ties Nest and Temporal SDK together and Nest users are already using it.
I don't know if there's that much value in having a sample here too.
I'd consider referencing https://github.com/KurtzL/nestjs-temporal in the README instead of having a full blown sample.

lorensr:
I’m happy to have more samples.

Haven’t looked at the difference between the two, but some options for
reducing maintenance burden are:

- using that lib in this sample
- referencing that lib in lieu of a sample (and perhaps contributing to
that lib if it doesn’t have everything the sample does)

In these cases it’s them updating / responding to issues/prs.

If those options aren’t easy or the best experience for users, I’m
personally down with sample

bergundy:
Let's explore the options and if we feel like we need a sample, let's add one.

rmnunes:
Hey guys,

I found the nestjs sample very useful. I was digging into the internet to find a good one. The only one is the referenced nestjs-temporal lib, which is quiet unfinished, if not an overkill, and lacks documentation. I think NestJs and Temporal is already a very nice combination, without external libraries.

It would be great to have maybe 3 samples from temporal team as reference.

xD

joebowbeer:
Also see lengthy NestJS discussion in https://github.com/temporalio/sdk-typescript/issues/483

I think my original hope was that someone would write a NestJS module for Temporal. Then if it received 10+ stars it could find its way into [awesome-nestjs](https://github.com/nestjs/awesome-nestjs) and then eventually it might become an official part of NestJS. Note that Zeebe is already listed in awesome-nestjs.

@rmnunes I would also like to see some simple NestJS samples included here.

bergundy:
I talked to @vkarpov15 today and asked him to continue working on this. Thanks everyone for showing interest.

vkarpov15:
@joebowbeer thanks for the link and the feedback. Do you have any other thoughts on this sample app?

I'd say the big blocker to having an officially supported NestJS integration is that we don't have anyone on the team who is sufficiently confident with NestJS to build and maintain one. This PR represents my first Nest app, the only Nest work I've done before is occasionally debugging Nest-related Mongoose issues.

Building a sample app is a good first step toward having NestJS docs and a NestJS integration, because we can discuss how to build an app with patterns that are idiomatic for both Temporal and Nest, without committing to maintaining an API yet.

ledniy:
@vkarpov15 Thank you for looking into it! If I understand correctly, there's no Activity used in the sample, which is the most important part from my point of view because most of the benefits of integrating Temporal with NestJS is to be able to use Nest infrastructure like dependency injection in the Activities for more complex things that just Hello World function

vkarpov15:
@ledniy @joebowbeer please check out f439a73, I added an example of creating a `WorkerModule` and `ActivitiesModule`. Does this seem like idiomatic NestJS? If not, can you please provide suggestions for improvement?

This example does start the Worker and WorkflowClient in the same process, which is a bit odd from a Temporal perspective, but seems to be more common in NestJS based on their [queues docs](https://docs.nestjs.com/techniques/queues#separate-processes).

ledniy:
@vkarpov15 looks good for me. In a real-world case, the client and worker would be split into different apps inside one monorepo (you can add an app and convert it to monorepo by `nest g app`)

rmnunes:
@vkarpov15 Thank you for putting effort to this sample! Amazing!
@ledniy I'm wondering in the case an Activity is injected with a dependency. If temporal will be able to resolve.

`@Injectable()
export class ActivitiesService {
  constructor(private db: DbService) {}
  async persist(value: number): Promise<unknown> {
    await this.db.save(1);
    const res = await axios.post('http://httpbin.org/post', { value });

    return res.data;
  }
}`

I was running some tests, and temporal is not able to find the module/service. I might be missing something. :/

`{
  "message": "Cannot read properties of undefined (reading 'save')",
  "source": "TypeScriptSDK",
  "stackTrace": "TypeError: Cannot read properties of undefined (reading 'persist')\n    at Activity.save[as fn] (..\\samples-typescript-vkarpov15-nestjs-counter\\nestjs-counter\\src\\activities\\activities.service.ts:9:19)",
  "cause": null,
  "applicationFailureInfo": {
    "type": "TypeError",
    "nonRetryable": false,
    "details": null
  }
}`

cheers :)

serhattanrikut:
@rmnunes 

I faced the same issue. Your solution works.. thanks.

ledniy:
@rmnunes I didn't have any problems injecting dependencies to ActivitiesService, and it should work the same way as any other provider in NestJS. The discovery approach seems like overhead for me

dbpolito:
@ledniy can you give an example on how you are using the injected dependency without https://github.com/temporalio/samples-typescript/pull/169#issuecomment-1204757472?

Because it seems `this` is referencing the `Activity` object instead of the service object...

```typescript
@Injectable()
export class Service1 {
  constructor(
    @Inject(AppService)
    private readonly appService: AppService,
  ) {}

  async run(value: string): Promise<string> {
    console.log(this);
    return `${this.appService.getHello()} ${value}`;
  }
}
```

ledniy:
@rmnunes @dbpolito my bad, I forgot that I did some fix to make it work, working example https://github.com/temporalio/samples-typescript/compare/vkarpov15/nestjs-counter...ledniy:samples-typescript:ledniy/nestjs-counter-di?expand=1

dbpolito:
@ledniy awesome, i was wondering if you were able solve this without binding, cool, i did the same thing here... 👍

nestjs-counter/src/counter-worker/counter-worker.providers.ts
```diff
@@ -0,0 +1,31 @@
...
+    provide: 'COUNTER_WORKER',
+    inject: [ActivitiesService],
+    useFactory: async (activitiesService: ActivitiesService) => {
+      Runtime.install({
+        logger: new DefaultLogger('WARN', (entry: LogEntry) =>
+          console.log(`[${entry.level}]`, entry.message),
+        ),
+      });
+
+      const activities = activitiesService as unknown as ActivityInterface;
```
bergundy, line 17: This shouldn’t be required anymore

bergundy: ActivityInterface was deprecated a while back

nestjs-counter/src/counter-workflow/counter-workflow.providers.ts
```diff
@@ -0,0 +1,35 @@
...
+  WorkflowExecutionAlreadyStartedError,
+} from '@temporalio/client';
+import { counterWorkflow } from '../temporal/workflows';
+import { taskQueue } from '../temporal/shared';
+
+export const counterWorkflowProviders = [
+  {
+    provide: 'COUNTER_WORKFLOW',
+    useFactory: async () => {
+      const client = new WorkflowClient();
```
bergundy, line 12: I’d show how to inject a connection here or even inject the client

nestjs-counter/src/temporal/activities.ts
```diff
@@ -0,0 +1,3 @@
+export type IActivities = {
```
bergundy, line 1: Don’t need this, can use the class directly

nestjs-counter/src/temporal/workflows.ts
```diff
@@ -0,0 +1,34 @@
...
+  defineSignal,
+  proxyActivities,
+  setHandler,
+} from '@temporalio/workflow';
+import { IActivities } from './activities';
+
+export const incrementSignal = defineSignal<[number]>('increment');
+export const getValueQuery = defineQuery<number>('getValue');
+
+const { persist } = proxyActivities<IActivities>({
```
bergundy, line 13: I’d use the class..

nestjs-counter/src/temporal/workflows.ts
```diff
@@ -0,0 +1,34 @@
...
+  startToCloseTimeout: '30 seconds',
+});
+
+export async function counterWorkflow(initial: number): Promise<void> {
```
bergundy, line 17: Not a good example workflow, history will grow indefinitely

nestjs-counter/src/temporal/workflows.ts
```diff
@@ -0,0 +1,34 @@
...
+  let value = initial;
+
+  setHandler(incrementSignal, async (val: number) => {
+    console.log('Increment', val);
+    if (!val) {
+      return;
+    }
+    value += val;
+
+    await persist(value);
```
bergundy, line 27: Async signal handlers are supported but not recommended. The workflow might complete or continue as new while the handler is running and you’ll end up losing progress

nestjs-counter/apps/worker/src/counter-worker/counter-worker.providers.ts
```diff
@@ -0,0 +1,31 @@
...
+          console.log(`[${entry.level}]`, entry.message),
+        ),
+      });
+
+      const activities = activitiesService as unknown as ActivityInterface;
+
+      const worker = await Worker.create({
+        workflowsPath: require.resolve('../temporal/workflows'),
+        taskQueue,
+        activities,
```
ledniy, line 22: @vkarpov15 pay attention that when activitiesService will have dependency injection, it will not work due to losing context. It was discussed above

vkarpov15:
@bergundy you're right that the current example is a bit contrived, I switched to using the exchange rates example from [this blog post](https://temporal.io/blog/caching-api-requests-with-long-lived-workflows). That means we can have an example that makes a realistic HTTP request, and also includes queries, without any async signals. Plus having the existing content helps clarify the example.

I'll work on fixing the rest of the comments later today :+1:

nestjs-counter/src/temporal/workflows.ts
vkarpov15: Fixed, I guess I just needed to upgrade SDK versions. With SDK v1.0.0-rc0, using the class led to TypeScript complaining "index signature for type 'string' is missing in type"

vkarpov15:
@ledniy @bergundy can you please take a look at the updated sample? I switched to using `@nestjs/axios`, NestJS monorepo, SDK v1.1.0, and did a bunch of cleanup. I think this sample now makes much more sense.

serhattanrikut:
> @rmnunes @dbpolito my bad, I forgot that I did some fix to make it work, working example https://github.com/temporalio/samples-typescript/compare/vkarpov15/nestjs-counter...ledniy:samples-typescript:ledniy/nestjs-counter-di?expand=1#diff-5ad4fdc317f0019f3bf1033a3f15e144390f0a132edc102ceb659c8c9c7b0520R18

cool...looks much better/easier. thank you

nestjs-counter/apps/server/src/app.module.ts
```diff
@@ -0,0 +1,8 @@
+import { Module } from '@nestjs/common';
```
bergundy, line 1: Can you fix these lint warnings? Looks like we need to add decorator support to eslint somehow..

nestjs-counter/apps/server/src/exchange-rates/exchange-rates.providers.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  WorkflowClient,
+  WorkflowExecutionAlreadyStartedError,
+} from '@temporalio/client';
+import { taskQueue } from '@app/shared';
+
+export const exchangeRatesProviders = [
+  {
+    provide: 'WORKFLOW_CLIENT',
+    useValue: new WorkflowClient(),
```
bergundy, line 10: Let's use the new `Client` here instead.

bergundy: I'd show how to inject a configured `Connection`, not just a `Client` that uses localhost.

nestjs-counter/apps/server/src/exchange-rates/exchange-rates.providers.ts
```diff
@@ -0,0 +1,37 @@
...
+  },
+  {
+    provide: 'EXCHANGE_RATES_WORKFLOW',
```
bergundy, line 13:
```suggestion
    provide: 'EXCHANGE_RATES_WORKFLOW_HANDLE',
```

?

nestjs-counter/apps/server/src/exchange-rates/exchange-rates.providers.ts
```diff
@@ -0,0 +1,37 @@
...
+      } catch (err) {
+        if (err instanceof WorkflowExecutionAlreadyStartedError) {
+          console.log('Reusing existing exchange rates workflow');
+          handle = await client.getHandle('exchange-rates');
+        } else {
+          throw err;
+        }
+      }
+
+      console.log('Started exchange rates workflow');
```
bergundy, line 32: Maybe move this log to after the `start` call?

nestjs-counter/apps/server/src/exchange-rates/exchange-rates.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+import { exchangeRatesProviders } from './exchange-rates.providers';
+
+describe('ExchangeRatesService', () => {
+  let app: TestingModule;
+  const rates = { AUD: 1.5 };
+
+  beforeAll(async () => {
+    const handleMock = {
+      query: jest.fn(() => rates),
+    };
```
bergundy, line 12: Why mock here? I'd rather show how to test a real app instead of relying on mocks

nestjs-counter/apps/worker/src/activities/activities.service.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+      imports: [HttpModule],
+      providers: [ActivitiesService],
+    }).compile();
+  });
+
+  describe('getExchangeRates', () => {
+    it('should return exchange rates', async () => {
+      const activitiesService = app.get(ActivitiesService);
+      const httpService = activitiesService.httpService;
+      jest.spyOn(httpService.axiosRef, 'get').mockImplementation(() =>
```
bergundy, line 19: This I can understand mocking

nestjs-counter/apps/worker/src/exchange-rates-worker/exchange-rates-worker.providers.ts
```diff
@@ -0,0 +1,33 @@
...
+export const exchangeRatesWorkerProviders = [
+  {
+    provide: 'EXCHANGE_RATES_WORKER',
+    inject: [ActivitiesService],
+    useFactory: async (activitiesService: ActivitiesService) => {
+      Runtime.install({
+        logger: new DefaultLogger('WARN', (entry: LogEntry) =>
+          console.log(`[${entry.level}]`, entry.message),
+        ),
+      });
```
bergundy, line 14: Why change the logger, especially don't love that we lose the timestamps here.
Also to follow the nestjs philosophy, `Runtime` should be a provider.

nestjs-counter/apps/worker/src/exchange-rates-worker/exchange-rates-worker.providers.ts
```diff
@@ -0,0 +1,33 @@
...
+
+      const activities = {
+        getExchangeRates:
+          activitiesService.getExchangeRates.bind(activitiesService),
+      };
+
+      const worker = await Worker.create({
+        workflowsPath: require.resolve('../temporal/workflows'),
```
bergundy, line 22: I'd show how to prove the native connection and more configuration like task queue / namespace.

nestjs-counter/apps/worker/src/main.ts
```diff
@@ -0,0 +1,8 @@
+import { NestFactory } from '@nestjs/core';
+import { ExchangeRatesWorkerModule } from './exchange-rates-worker/exchange-rates-worker.module';
+
+async function bootstrap() {
+  const app = await NestFactory.create(ExchangeRatesWorkerModule);
+  await app.listen(3001);
```
bergundy, line 6: What is this listening for?

nestjs-counter/apps/worker/src/temporal/workflows.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    // Filter INFO log messages for clearer test output
+    Runtime.install({ logger: new DefaultLogger('WARN') });
+    env = await TestWorkflowEnvironment.create();
+    const worker = await Worker.create({
+      connection: env.nativeConnection,
+      taskQueue,
+      workflowsPath: require.resolve('./workflows.ts'),
+      activities,
+    });
```
bergundy, line 30: I'd show how to integrate the test environment with nestjs so you can use the worker provider instead of rewriting all of this in the test.

nestjs-counter/apps/worker/src/temporal/workflows.ts
```diff
@@ -0,0 +1,43 @@
...
+import {
+  getExchangeRatesQuery,
+  ExchangeRates,
+} from '../../../../libs/shared/src';
+
+const { getExchangeRates } = proxyActivities<ActivitiesService>({
+  startToCloseTimeout: '1 minute',
+});
+
+const maxIterations = 10000;
```
bergundy, line 19: Use `workflowInfo().historyLength` instead of counting iterations.

nestjs-counter/apps/worker/src/temporal/workflows.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  continueAsNew,
+  proxyActivities,
+  setHandler,
+  sleep,
+} from '@temporalio/workflow';
+import { ActivitiesService } from '../activities/activities.service';
```
bergundy, line 7:
```suggestion
import type { ActivitiesService } from '../activities/activities.service';
```

nestjs-counter/apps/worker/src/temporal/workflows.ts
```diff
@@ -0,0 +1,43 @@
...
+  ExchangeRates,
+} from '../../../../libs/shared/src';
+
+const { getExchangeRates } = proxyActivities<ActivitiesService>({
+  startToCloseTimeout: '1 minute',
+});
+
+const maxIterations = 10000;
+
+export async function exchangeRatesWorkflow(
```
bergundy, line 21: We're not bringing anything interesting from NextJS to workflows, should we try to use some of their module / provider system to construct workflows?

nestjs-counter/apps/worker/src/temporal/workflows.ts
```diff
@@ -0,0 +1,43 @@
...
+  storedRates: ExchangeRates | null = null,
```
bergundy, line 22: nit:

```suggestion
  storedRates?: ExchangeRate,
```

bergundy: The the query can return undefined instead of null too

nestjs-counter/nest-cli.json
```diff
@@ -0,0 +1,40 @@
...
+    "shared": {
+      "type": "library",
+      "root": "libs/shared",
+      "entryFile": "index",
+      "sourceRoot": "libs/shared/src",
+      "compilerOptions": {
+        "tsConfigPath": "libs/shared/tsconfig.lib.json"
+      }
+    }
+  }
```
bergundy, line 39: Why do we need 3 projects? What does this buy us?

nestjs-counter/apps/server/src/app.module.ts
vkarpov15: Where are the lint warnings? It looks like lint succeeds in CI...

nestjs-counter/apps/server/src/exchange-rates/exchange-rates.spec.ts
vkarpov15: 2 reasons:

1) I had a fair bit of trouble getting the test env to work with Jest and Nest.
2) Strict unit testing (stub everything, only test the code you're looking at) seems to be more idiomatic for Nest.

I don't hold very strong opinions here. Personally, I'd prefer to test a real app in this case. But I'd also like to keep this sample as idiomatic as reasonably possible for Nest. So I'd love to hear any thoughts or feedback.

nestjs-counter/apps/worker/src/main.ts
vkarpov15: As far as I can tell, there's no way to start a Nest app without starting an HTTP server. So this just starts an HTTP server that does nothing.

nestjs-counter/apps/worker/src/temporal/workflows.ts
vkarpov15: I'm not sure how well using modules and providers in Workflows would work, given that we transpile and bundle Workflows. What if some part of NestJS uses, or starts using, some Node built-ins or something else that is non-deterministic? Plus, there isn't much we can inject, other than maybe `setHandler()`.

nestjs-counter/apps/server/src/app.module.ts
bergundy: image

nestjs-counter/apps/worker/src/exchange-rates-worker/exchange-rates-worker.providers.ts
bergundy: Still think this is relevant.

nestjs-counter/apps/worker/src/main.ts
bergundy: I'd add a comment in the code

nestjs-counter/apps/worker/src/temporal/workflows.spec.ts
bergundy: Comment still applies.

nestjs-counter/apps/worker/src/temporal/workflows.ts
bergundy: Having 0 xp with nestjs I figure we can leave this open for now.

nestjs-counter/nest-cli.json
bergundy: 🤷

bergundy (approved):
At this point I feel like the code is good enough to merge.
I still have some unaddressed comments but I won't block the PR anymore.

nestjs-counter/apps/worker/src/exchange-rates-worker/exchange-rates-worker.providers.ts
```diff
@@ -0,0 +1,27 @@
...
+        taskQueue,
+        activities,
+      });
+
+      worker.run();
```
ledniy, line 21: do we need `await` here?

lorensr:
@vkarpov15 Looking forward to getting this sample in ☺️ I enabled auto merge—could you get the tests to pass?

PR #171, opened 2022-07-14
Make encryption example compatible with other SDKs.

robholland:
## What was changed

Switched away from crypto-async library which is no longer maintained and 3 years stale. Made the crypto and payload marshalling compatible with our other SDK encryption examples.

## Why?

This makes for a better starting point for users wishing to use encryption in a polyglot setup.

## Checklist

1. Closes 

2. How was this tested:

Tested locally using Typescript Codec Server against Typescript and Go SDK encrypted payloads.

3. Any docs updates needed?


bergundy (commented):
The reason for using that crypto library was to avoid blocking the main nodejs thread.
I suggest we run codec in a thread pool like this one: https://github.com/piscinajs/piscina

robholland:
Not sure if you mean adjust this sample or adjust the SDK code to use a thread pool. Either way I think that's better done by someone with more Typescript-fu than me.

lorensr:
If we limit support to Node 15+, I believe the Web Crypto API uses threadpool:

https://nodejs.org/api/webcrypto.html#encryption-and-decryption

image

https://github.com/nodejs/node/blob/42ad967d68137df1a80a877e7b5ad56403fc157f/src/crypto/README.md

bergundy:
Leaving this up to @lorensr to fix, thanks @robholland

lorensr:
@robholland looks like Web Crypto API doesn't support 16-bit tag length. Could we update all samples to one of the recommended lengths, like 128? 

image

If not, then I'll code a thread pool instead of using Web Crypto API.

robholland:
The TAG_LENGTH variable in the PR is in bytes, not bits. 16 bytes is 128 bits.

lorensr:
Oh great, thanks!

cretz:
Note I may have copied some of this impl a bit when I wrote https://github.com/temporalio/samples-python/tree/main/encryption. Feel free to review that one if you want.

lorensr:
Tested via:

- Change `encryption/client.ts` type:

```
  const handle = await client.start('Workflow', {
```

- Comment out this line in Go [encryption sample](https://github.com/temporalio/samples-go/tree/main/encryption):

```
	// 	codecs = append(codecs, converter.NewZlibCodec(converter.ZlibCodecOptions{AlwaysEncode: true}))
```	

Run `go run worker/main.go` and then `npm run workflow`. Output:

```
> temporal-encryption@0.1.0 workflow
> ts-node src/client.ts

Started workflow my-business-id-57fb8cfd-8276-46f0-ae2f-84bf1b7ebab3
Decrypting payload.data: <Buffer 0f 35 6c 35 97 13 03 80 9b 87 9a 20 c8 d8 9c 44 5b 39 34 05 1b d2 32 d3 f9 a1 da 88 8f ce c0 ad 2a e5 6e a7 1b a4 33 2d f5 99 27 95 b8 d3 c0 a3 0e 5c ... 44 more bytes>
Hello Alice: Private message for Bob.!
```

bergundy:
Thanks for getting this merged, @lorensr!

PR #167, opened 2022-07-07
Deprecate temporalio

lorensr:
Ran 

```
npm rm temporalio; npm install @temporalio/{client,worker,workflow,activity}; npx sort-package-json
```

in most samples. monorepo has separate packages, and some samples import from common. (Some of which can be removed after https://github.com/temporalio/sdk-typescript/pull/736 is released)

Fixes https://github.com/temporalio/sdk-typescript/issues/727

PR #173, opened 2022-07-15
Fix upgrade script

lorensr:
Missed this comment: https://github.com/temporalio/samples-typescript/pull/167#discussion_r921598313


bergundy (approved):
Might as well make a script out of it but 🤷

PR #178, opened 2022-07-26
Update versions and README instructions

lorensr:


bergundy (approved):
**nit**: Let's remove common if not strictly required, WDYT?

PR #182, opened 2022-07-29
Update to v1.0.1

lorensr:


lorensr:
Cons of changing to `^1.0.0`? Pro: don't need PRs like this

bergundy (commented):
Let's do `^1.0.0` we should be able to rely on semver.

bergundy:
Only reason the version was fixed is because of the rc suffix.

bergundy:
Replaced with #184

PR #190, opened 2022-08-20
Update for v1.1.0

lorensr:


.scripts/upgrade-versions.mjs
```diff
@@ -21,6 +21,15 @@ for (const sample of STORED_SAMPLES) {
...
     }
   }
+  for (const [dep, version] of Object.entries(content.devDependencies)) {
+    if (dep.startsWith('@temporalio/')) {
+      if (version !== versionToSet) {
+        content.devDependencies[dep] = versionToSet;
+        console.log(`Updated version for ${sample} devDependency ${dep} from ${version} to ${versionToSet}`);
+        changed = true;
+      }
+    }
```
bergundy, line 31: nit: move to function

bergundy (approved):

.scripts/upgrade-versions.mjs
lorensr: The internet has been telling me lately to stop drying every single thing, so I resisted the urge 😄 Will change

PR #197, opened 2022-09-17
Fix error: Workflow referenced an unregistered external sink

lorensr:


logging-sinks/src/worker.ts
```diff
@@ -1,9 +1,10 @@
 // @@@SNIPSTART typescript-logger-sink-worker
-import { Worker, InjectedSinks } from '@temporalio/worker';
-import { LoggerSinks } from './workflows';
+import { defaultSinks, InjectedSinks, LoggerSinks, Worker } from '@temporalio/worker';
+import { CustomLoggerSinks } from './workflows';
 
 async function main() {
-  const sinks: InjectedSinks<LoggerSinks> = {
+  const sinks: InjectedSinks<LoggerSinks & CustomLoggerSinks> = {
```
lorensr, line 6: @bergundy Alternatively, SDK could export `LoggerSinks` from `@temporalio/workflow`, and then user can do 

```ts
export type AllSinks = LoggerSinks & MySinks
```

in workflow code, and use it both in workflow and here in worker code.

lorensr:
Will compile once https://github.com/temporalio/sdk-typescript/pull/871 is merged & published

logging-sinks/src/worker.ts
bergundy: Good idea

bergundy (approved):
I think the instrumentation sample needs an update too

instrumentation/src/worker.ts
```diff
@@ -50,28 +59,24 @@ async function main() {
...
-        fn(workflowInfo, message, meta) {
-          workflowLogger.child(workflowInfo).warn(message, meta);
-        },
-      },
+  const workflowLogger: Logger = {
+    log(level: LogLevel, message: string, meta?: LogMetadata) {
+      workflowWinstonLogger.log(message, meta);
+    },
+    trace(message: string, meta?: LogMetadata) {
+      workflowWinstonLogger.verbose(message, meta);
```
lorensr, line 67: no trace option

bergundy (approved):
Nice!

PR #198, opened 2022-09-20
v1.3

lorensr:


activities-examples/src/workflows.test.ts
```diff
@@ -12,19 +12,15 @@ beforeAll(async () => {
     logger: new DefaultLogger('WARN', (entry: LogEntry) => console.log(`[${entry.level}]`, entry.message)),
   });
 
-  testEnv = await TestWorkflowEnvironment.create({
-    testServer: {
-      stdio: 'inherit',
-    },
```
lorensr, line 18: @bergundy I forget—is this now default behavior, or is there another equiv option?

bergundy: There’s no output from the test server so this doesn’t matter in practice, for temporalite we allow configuring the log level

bergundy (approved):
I’m debating whether to change the samples to use the new Client instead

lorensr:
I don't think it's important. It's simple to use `WorkflowClient` when that's the only client you use. In cases where you're using multiple clients, I'd use `Client`.

PR #199, opened 2022-09-29
Update for v1.4

lorensr:


ejson/src/ejson-payload-converter.ts
```diff
@@ -1,13 +1,7 @@
...
-  EncodingType,
-  errorMessage,
-  METADATA_ENCODING_KEY,
-  Payload,
-  PayloadConverterWithEncoding,
-  str,
-  u8,
-} from '@temporalio/common';
+import { EncodingType, METADATA_ENCODING_KEY, Payload, PayloadConverterWithEncoding } from '@temporalio/common';
+import { errorMessage } from '@temporalio/common/lib/type-helpers';
```
bergundy, line 3: Let's not use this internal API

ejson/src/ejson-payload-converter.ts
```diff
@@ -1,13 +1,7 @@
...
+import { decode, encode } from '@temporalio/common/lib/encoding';
```
bergundy, line 4: As mentioned on slack, let's use an external lib for this.

bergundy (commented):
Look below at the deprecated uses of `u8` and `str`, we should replace those with TextEncoder calls (requires `lib: 'dom'` in `tsconfig.json`).

hello-world/src/workflows.ts
```diff
@@ -9,6 +9,11 @@ const { greet } = proxyActivities<typeof activities>({
 
 /** A workflow that simply calls an activity */
 export async function example(name: string): Promise<string> {
-  return await greet(name);
+  const minuteP = sleep('1m');
```
bergundy, line 12: Assuming this wasn’t supposed to be committed

lorensr: thanks!

vkarpov15 (approved):

PR #204, opened 2022-11-02
Use fastestsmallesttextencoderdecoder

lorensr:
Any ideas on this webpack behavior? When I log from the client, I get:

image

but in the workflow, it's an empty object `{}`:

image

Similar message when importing default

```ts
import encoder from 'fastestsmallesttextencoderdecoder';
```

Repro:

```
git checkout encoding-types
cd ejson
npm i
npm start
npm run workflow
```

### DIFF
--- a/ejson/package.json
+++ b/ejson/package.json
@@ -32,6 +32,7 @@
   "devDependencies": {
     "@tsconfig/node16": "^1.0.0",
     "@types/ejson": "^2.1.3",
+    "@types/fastestsmallesttextencoderdecoder": "^1.0.0",
     "@types/node": "^16.11.43",
     "@types/uuid": "^8.3.4",
     "@typescript-eslint/eslint-plugin": "^5.0.0",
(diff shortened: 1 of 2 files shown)


mjameswh:
That package provides a ["browser" export](https://github.com/anonyco/FastestSmallestTextEncoderDecoder/blob/master/package.json#L54), which gets picked up by Webpack instead of the NodeJS variant. We can see this in the Webpack's output at build time:

Capture d’écran, le 2022-11-07 à 18 09 14

Note that EncoderDecoderTogether.min.js is the name of the browser variant of the package, rather than the NodeJS variant. Having a look at the source code of that file, it appears that it expose itself through globals (`window.TextDecoder` and `window.TextEncoder`) instead of `module.exports`, which explains why you get an empty object for that module.

Now, just for testing, I succeeded forcing webpack to pick the correct variant using the following code:
```
    bundlerOptions: {
      webpackConfigHook(config: WebpackConfiguration) {
        config!.resolve!.alias = {
          ...config!.resolve!.alias,
          fastestsmallesttextencoderdecoder: `${dirname(__dirname)}/node_modules/fastestsmallesttextencoderdecoder/NodeJS/EncoderAndDecoderNodeJS.min.js`
        };
        return config;
      },
    }
```

... Turns out I get another error, because it is trying to access `global["Buffer"];` and can't find it... Maybe you should consider some other library?

bergundy:
Let's reexport the SDK inlined implementation of this library.
I was hoping we could reduce the API surface but it doesn't seem worth it right now.
We can close this PR.

PR #212, opened 2022-11-29
add snipsync comment wrappers

kimschles:
## What was changed
I added snipsync comment wrappers to the `package.json` file in the hello-world project.

## Why?
I am writing a tutorial using the Hello World project, and I would like snipsync to pull in the `package.json` file.

### DIFF
--- a/hello-world/package.json
+++ b/hello-world/package.json
@@ -1,3 +1,4 @@
+// @@@SNIPSTART typescript-package-json
 {
   "name": "temporal-hello-world",
   "version": "0.1.0",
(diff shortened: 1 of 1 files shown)


lorensr:
JSON doesn't support any type of comments, including lines starting with `//`, so snipsync won't work here (see [CI failure](https://github.com/temporalio/samples-typescript/actions/runs/3571377863/jobs/6003192315).

lorensr:
We may be able to add a feature to snipsync like this: https://github.com/temporalio/snipsync/issues/58

kimschles:
Thanks for the info, and thanks for opening that issue, @lorensr. For now, I'll close this PR and keep an eye on https://github.com/temporalio/snipsync/issues/58.

PR #213, opened 2022-12-08
chore: Update samples to SDK 1.5.2

mjameswh:
## What was changed

- Updated all @temporalio/* dependencies to `^1.5.2`
- Replaced some deprecated or legacy syntaxes by their recommended alternatives
- Updated lerna to 6.1


bergundy (approved):
I would do a review of the docs and see if anything that references these snippets needs to be updated

PR #220, opened 2022-12-16
Run separate workflows to test monorepo-folders, nestjs-counter samples

vkarpov15:
## What was changed

Added a separate workflow to lint and test monorepo-folders and nestjs-counter because of the out of space issues with #169 .

## Why?

Tests started running out of disk space on Windows because installing a separate copy of the TypeScript SDK, along with all the other deps, for every single sample means that disk space use grows linearly with the number of samples. Eventually, with enough samples, we'll run out of disk space.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


.github/workflows/ci.yml
```diff
@@ -2,8 +2,6 @@ name: CI
 
 on:
   push:
-    branches:
-      - main
```
lorensr, line 6: Why run on push to all branches?

lorensr (approved):
Some lint warnings at bottom of Files changed tab

.github/workflows/ci.yml
vkarpov15: I primarily made this change for testing, so the tests would actually run on my changes before I put in a PR. I'll revert it.

vkarpov15:
The lint warnings are due to lack of decorator support, I'll look into figuring out how to add decorator support for eslint in the nestjs-counter example. Decorators are pretty important in Nest.

PR #221, opened 2023-01-01
Add schedules sample

### LINKED ISSUE #215: [Feature Request] Add Schedules usage example
### Description

TS SDK 1.5.0 added support for Schedules. Need to come up with a sample on how to use.

lorensr:
- Use Temporal CLI (not very publicized yet but works fine—think it's okay to merge a little early here)
- Add schedules sample

Fixes #215


vkarpov15 (approved):
LGTM :+1:

lorensr:
hah 😄

image

mjameswh:
> hah 😄
> 
> image

Ouch. I feel your pain. 😄

.shared/.post-create
```diff
@@ -1,13 +1,11 @@
...
+  Other: Download and extract the latest release from https://github.com/temporalio/cli/releases/latest
 
 Start Temporal Server:
 
-  {cyan git clone} https://github.com/temporalio/docker-compose.git temporal-server
-  {cyan cd} temporal-server/
-  {cyan docker-compose up}
+  {cyan temporal server start-dev}
 
 Use Node version 14+:
```
mjameswh, line 10: While you're at this... Could you please update this to 14.18? Might even recommend 16+

schedules/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "execMap": {
+      "ts": "ts-node"
+    },
+    "ext": "ts",
+    "watch": [
+      "src"
+    ]
+  },
+  "dependencies": {
+    "@temporalio/activity": "^1.0.0",
```
mjameswh, line 27: These should be "^1.5.2"

lorensr: Why did this change? I remember changing everything to `^1.0.0` at Roey's suggestion

schedules/src/pause-schedule.ts
```diff
@@ -0,0 +1,19 @@
...
+  });
+
+  const handle = client.schedule.getHandle('sample-schedule');
+  await handle.pause();
+
+  console.log(`Schedule is now paused.`);
+}
+
+run()
+  .then(() => process.exit(0))
```
mjameswh, line 15: I'm hesitant about all `exit(0)`...

lorensr: Won't need anymore given your dep pin fix, will remove

schedules/package.json
mjameswh: Do you remember @bergundy argument for this suggestion?

IMHO: With `^1.0.0`, NPM will happily use whatever old version that the user has in his NPM cache. Many examples could therefore fail to run in user environment. Keeping version number in samples up-to-date means we silently push users toward updating. My intention would be to automatically update this number from sdk-typescript's release CI.

lorensr: Probably so we didn't have to keep doing the manual script. If it's automatic, sounds good!

PR #222, opened 2023-01-01
Mutex workflow sample

### LINKED ISSUE #214: [Feature Request] Add a mutex workflow sample
Same as this go sample: https://github.com/temporalio/samples-go/blob/main/mutex/mutex_workflow.go
Rough implementation (untested):

```ts
import * as workflow from '@temporalio/workflow';

interface LockRequest {
  initiatorId: string;
  timeoutMs: number;
}

interface LockResponse {
  releaseSignalName: string;
}

const lockRequestSignal = workflow.defineSignal<[LockRequest]>('lock-requested');
const lockAcquiredSignal = workflow.defineSignal<[LockResponse]>('lock-acquired');


async function lockWorkflow(requests = Array<LockRequest>()) {
  workflow.setHandler(lockRequestSignal, (req) => {
    requests.push(req);
  });
  while (workflow.workflowInfo().historyLength < 2000) {
    await workflow.condition(() => requests.length > 0);
    const req = requests.shift();
    const handle =  workflow.getExternalWorkflowHandle(req.initiatorId);
    const releaseSignalName = workflow.uuid4();
    await handle.signal(lockAcquiredSignal, { releaseSignalName });
    released = false;
    setHandler(workflow.defineSignal(releaseSignalName), () => {
      released = true;
    });
    await workflow.condition(() => released, req.timeoutMs);
  }
  // carry over any pending requests to the next execution
  await workflow.continueAsNew<typeof lockWorkflow>(requests);
}
```

vkarpov15:
## What was changed

Added a mutex workflow sample as described in #214 

## Why?

This sample is a good demo of inter-workflow communication - shows you can signal and query a Workflow from another Workflow, without either Workflow being a child of the other. I don't think we have another sample that shows this.

## Checklist

1. Closes #214

2. How was this tested:

3. Any docs updates needed?


mutex/src/workflows.ts
```diff
@@ -0,0 +1,66 @@
...
+  workflow.setHandler(lockRequestSignal, (req: LockRequest) => {
+    requests.push(req);
+  });
+  workflow.setHandler(currentWorkflowIdQuery, () => currentWorkflowId);
+  while (workflow.workflowInfo().historyLength < 2000) {
+    await workflow.condition(() => requests.length > 0);
+    const req = requests.shift();
+    if (req === undefined) {
+      continue;
+    }
```
lorensr, line 28: when would this be true?

mutex/src/workflows.ts
```diff
@@ -0,0 +1,66 @@
...
+    currentWorkflowId = req.initiatorId;
+    const handle = workflow.getExternalWorkflowHandle(req.initiatorId);
+    const releaseSignalName = workflow.uuid4();
+    await handle.signal(lockAcquiredSignal, { releaseSignalName });
```
lorensr, line 32:
```suggestion
    const workflowRequestingLock = workflow.getExternalWorkflowHandle(req.initiatorId);
    const releaseSignalName = workflow.uuid4();
    await workflowRequestingLock.signal(lockAcquiredSignal, { releaseSignalName });
```

mutex/src/workflows.ts
```diff
@@ -0,0 +1,66 @@
+import * as workflow from '@temporalio/workflow';
```
lorensr, line 1: curious why namespace import vs named import. personally, i don't like having to remember while reading the code that `workflow` is a package—my brain assumes it's an object that's being operated on. and it feels like a lot of `workflow.`s to read/skip over. to me, those things are more important than the reader having to know/remember that all these functions are imported vs defined in this file.

mutex/src/workflows.ts
```diff
@@ -0,0 +1,66 @@
...
+  const { workflowId } = workflow.workflowInfo();
+
+  let releaseSignalName: string | null = null;
+  workflow.setHandler(lockAcquiredSignal, (lockResponse: LockResponse) => {
+    releaseSignalName = lockResponse.releaseSignalName;
+  });
+  workflow.setHandler(hasLockQuery, () => !!releaseSignalName);
+
+  await handle.signal(lockRequestSignal, { timeoutMs: lockTimeoutMs, initiatorId: workflowId });
+  await workflow.condition(() => !!releaseSignalName);
```
lorensr, line 56:
```suggestion
  const hasLock = () => !!releaseSignalName
  workflow.setHandler(hasLockQuery, hasLock);

  await handle.signal(lockRequestSignal, { timeoutMs: lockTimeoutMs, initiatorId: workflowId });
  await workflow.condition(hasLock);
```

mutex/src/workflows.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  await workflow.sleep(sleepForMs);
+
+  if (!releaseSignalName) {
+    return;
+  }
```
lorensr, line 62: to protect against another signal being sent with `{ releaseSignalName: <falsy> }`? think okay to simplify and assume good actors.

mutex/src/workflows.ts
```diff
@@ -0,0 +1,66 @@
...
+interface LockResponse {
+  releaseSignalName: string;
+}
+
+export const currentWorkflowIdQuery = workflow.defineQuery<string | null>('current-workflow-id');
+export const hasLockQuery = workflow.defineQuery<boolean>('hasLock');
+export const lockRequestSignal = workflow.defineSignal<[LockRequest]>('lock-requested');
+export const lockAcquiredSignal = workflow.defineSignal<[LockResponse]>('lock-acquired');
+
+export async function lockWorkflow(requests = Array<LockRequest>()) {
```
lorensr, line 17:
```suggestion
export async function lockWorkflow(requests = Array<LockRequest>()): Promise<void> {
```

mutex/src/workflows.ts
```diff
@@ -0,0 +1,66 @@
...
+      released = true;
+    });
+    await workflow.condition(() => released, req.timeoutMs);
+    currentWorkflowId = null;
+  }
+  // carry over any pending requests to the next execution
+  await workflow.continueAsNew<typeof lockWorkflow>(requests);
+}
+
+export async function testLockWorkflow(lockWorkflowId: string, sleepForMs = 500, lockTimeoutMs = 1000) {
```
lorensr, line 44:
```suggestion
export async function testLockWorkflow(lockWorkflowId: string, sleepForMs = 500, lockTimeoutMs = 1000): Promise<void> {
```

lorensr, line 44: maybe name it `taskProcessor` or something indicating what you'd use a lock for

lorensr (approved):

mutex/src/workflows.ts
vkarpov15: Never. This was just to prevent TS from complaining that `req` is potentially undefined on subsequent lines.

mutex/src/workflows.ts
vkarpov15: Copy/pasted and didn't change it. I'll switch to named imports.

mutex/src/workflows.ts
vkarpov15: No, this is just to appease TypeScript, because TS thinks that `releaseSignalName` may be null.

mutex/src/workflows.ts
vkarpov15: I think `taskProcessor` is a little too generic, doesn't provide enough info about what this Workflow does. I'm fine with changing this, but I haven't been able to come up with a name that I like better than `testLockWorkflow`.

mutex/.post-create
```diff
@@ -0,0 +1,20 @@
...
+
+Use Node version 14+:
+
+  Mac: {cyan brew install node@16}
+  Other: https://nodejs.org/en/download/
+
+Then, in the project directory, using two other shells, run these commands:
+
+  {cyan npm run start.watch}
+  {cyan npm run workflow}
```
bergundy, line 20: You forgot to update the instructions here

mutex/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "watch": [
+      "src"
+    ]
+  },
+  "dependencies": {
+    "@temporalio/activity": "^1.5.2",
+    "@temporalio/client": "^1.5.2",
+    "@temporalio/worker": "^1.5.2",
+    "@temporalio/workflow": "^1.5.2",
+    "uuid": "8.x"
```
bergundy, line 29: nit: can we use `crypto.randomUUID` instead of adding this dependency?

mutex/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@types/uuid": "8.x",
+    "@typescript-eslint/eslint-plugin": "^5.0.0",
+    "@typescript-eslint/parser": "^5.0.0",
+    "eslint": "^7.32.0",
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-deprecation": "^1.2.1",
+    "mocha": "10.2.0",
+    "nodemon": "^2.0.12",
+    "ts-node": "^10.8.1",
+    "typescript": "^4.4.2"
```
bergundy, line 45: nit: Can we upgrade the minimum typescript version? (can run ncu in this folder to update all deps).

mutex/src/start-lock-workflow.ts
```diff
@@ -0,0 +1,21 @@
+import { Connection, Client } from '@temporalio/client';
+import { lockWorkflow } from './workflows';
+import { v4 as uuidv4 } from 'uuid';
+
+async function run() {
+  const connection = await Connection.connect();
+
+  const client = new Client({ connection });
+
+  const workflowId = 'lock-' + uuidv4();
```
bergundy, line 10: There's no need to explicitly start this workflow, in a real world example you'd use signalWithStart (from a local activity).
Please use the Go sample as reference: https://github.com/temporalio/samples-go/blob/e7f07468217806f57b22205692c4d70608b3e68b/mutex/mutex_workflow.go#L45

mutex/src/test/workflows.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+describe('lock workflow', function () {
+  let runPromise: Promise<void>;
+  let worker: Worker;
+  let lockWorkflowHandle: WorkflowHandle<typeof lockWorkflow>;
+  let env: TestWorkflowEnvironment;
+
+  before(async function () {
+    this.timeout(10000);
+    Runtime.install({ logger: new DefaultLogger('WARN') });
```
bergundy, line 19: No need to customize the logger anymore.
The default INFO level is fine IMHO, I changed the logger to use `process.stderr.write` directly instead of `console.error` and `jest` no longer warns about this.

I'd prefer to get the INFO messages and get a feel for what's going on even if the cost is more verbose output in tests.

Leaving it up to your consideration, run it and get a feel for it.

mutex/src/test/workflows.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    env = await TestWorkflowEnvironment.create();
+
+    worker = await Worker.create({
+      connection: env.nativeConnection,
+      workflowsPath: require.resolve('../workflows'),
+      taskQueue,
+    });
+
+    runPromise = worker.run();
```
bergundy, line 28: nit: use a global bundle and worker per test with `runUntil`.
It's better for error handling an propagating fatal errors out of the worker.

mutex/src/workflows.ts
```diff
@@ -0,0 +1,76 @@
...
+  initiatorId: string;
+  timeoutMs: number;
+}
+
+interface LockResponse {
+  releaseSignalName: string;
+}
+
+export const currentWorkflowIdQuery = defineQuery<string | null>('current-workflow-id');
+export const hasLockQuery = defineQuery<boolean>('hasLock');
```
bergundy, line 23: Please use consistent naming (`has-lock`)

mutex/src/workflows.ts
```diff
@@ -0,0 +1,76 @@
...
+export const lockRequestSignal = defineSignal<[LockRequest]>('lock-requested');
+export const lockAcquiredSignal = defineSignal<[LockResponse]>('lock-acquired');
+
+export async function lockWorkflow(requests = Array<LockRequest>()): Promise<void> {
+  let currentWorkflowId: string | null = null;
+  setHandler(lockRequestSignal, (req: LockRequest) => {
+    requests.push(req);
+  });
+  setHandler(currentWorkflowIdQuery, () => currentWorkflowId);
+  while (workflowInfo().historyLength < 2000) {
```
bergundy, line 33: nit: extract `2000` to constant.

mutex/src/workflows.ts
bergundy: Yeah, TS can't know that the array is empty here.
You might want to add a comment explaining why this is here.

mutex/src/workflows.ts
```diff
@@ -0,0 +1,76 @@
...
+    await workflowRequestingLock.signal(lockAcquiredSignal, { releaseSignalName });
+    let released = false;
+    setHandler(defineSignal(releaseSignalName), () => {
+      released = true;
+    });
+    await condition(() => released, req.timeoutMs);
+    currentWorkflowId = null;
+  }
+  // carry over any pending requests to the next execution
+  await continueAsNew<typeof lockWorkflow>(requests);
```
bergundy, line 51: nit: Since we're using `signalWithStart` there's no need to CAN if there are no pending requests at this point.

mutex/src/workflows.ts
```diff
@@ -0,0 +1,76 @@
...
+    releaseSignalName = lockResponse.releaseSignalName;
+  });
+  const hasLock = () => !!releaseSignalName;
+  setHandler(hasLockQuery, hasLock);
+
+  // Acquire the lock
+  await handle.signal(lockRequestSignal, { timeoutMs: lockTimeoutMs, initiatorId: workflowId });
+  await condition(hasLock);
+
+  // Critical path
```
bergundy, line 70: Please add some log statements like the Go sample so it's easier to grasp what's going on when running this sample.

mutex/src/workflows.ts
```diff
@@ -0,0 +1,76 @@
...
+export async function testLockWorkflow(lockWorkflowId: string, sleepForMs = 500, lockTimeoutMs = 1000): Promise<void> {
+  const handle = getExternalWorkflowHandle(lockWorkflowId);
+
+  const { workflowId } = workflowInfo();
+
+  let releaseSignalName = '';
+  setHandler(lockAcquiredSignal, (lockResponse: LockResponse) => {
+    releaseSignalName = lockResponse.releaseSignalName;
+  });
+  const hasLock = () => !!releaseSignalName;
```
bergundy, line 63: This query is likely to be inconsistent, I'd suggest removing it altogether and relying on the lock workflow's query alone.

mutex/src/test/workflows.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  it('handles locking and unlocking', async function () {
+    const testWorkflowId = 'test-' + uuidv4();
+    const testWorkflowHandle = await env.workflowClient.start(testLockWorkflow, {
+      taskQueue,
+      workflowId: testWorkflowId,
+      args: [lockWorkflowHandle.workflowId],
+    });
+
+    let hasLock = await testWorkflowHandle.query(hasLockQuery);
+    assert.ok(!hasLock);
```
bergundy, line 54: This seems racy and has a chance of failing on an overloaded machine, you should get rid of this.

mutex/src/test/workflows.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    const currentWorkflowId = await lockWorkflowHandle.query(currentWorkflowIdQuery);
+    assert.equal(currentWorkflowId, testWorkflowId);
+
+    await testWorkflowHandle.result();
+
+    hasLock = await testWorkflowHandle.query(hasLockQuery);
+    assert.ok(!hasLock);
+  });
+
+  it('errors out if waiting for lock times out', async function () {
```
bergundy, line 70: The title says "errors out" but I don't see any errors caught in this test.

bergundy (commented):
Thanks @vkarpov15, I left some comments.

mutex/src/workflows.ts
lorensr: is casting better?

```
const req = requests.shift() as LockRequest; // this will never be undefined
```

mutex/src/workflows.ts
lorensr: Maybe `oneAtATimeWorkflow` that calls an activity `useAPIThatCantBeCalledInParallel()` 😄

mutex/package.json
vkarpov15: Will switch to nanoid to be consistent with other samples. Nanoid is nice because it generates a much shorter id than a uuidv4, which is better for an example app.

mutex/src/test/workflows.test.ts
vkarpov15: I think it is better DX to suppress the additional log info in tests, at least for samples. Just having the output from the test runner helps new devs feel that nothing went wrong. The below output can be a bit jarring and make inexperienced users think an error occurred.

```
2023-01-05T15:50:06.637Z [INFO] Creating worker {
  options: {
    namespace: 'default',
    identity: '142091@defiant',
    shutdownGraceTime: '10s',
    maxConcurrentActivityTaskExecutions: 100,
    maxConcurrentLocalActivityExecutions: 100,
    enableNonLocalActivities: true,
    maxConcurrentWorkflowTaskExecutions: 100,
    stickyQueueScheduleToStartTimeout: '10s',
    maxHeartbeatThrottleInterval: '60s',
    defaultHeartbeatThrottleInterval: '30s',
    isolateExecutionTimeout: '5s',
    workflowThreadPoolSize: 8,
    maxCachedWorkflows: 261,
    enableSDKTracing: false,
    showStackTraceSources: false,
    debugMode: false,
    interceptors: { activityInbound: [Array], workflowModules: [Array] },
    sinks: { defaultWorkerLogger: [Object] },
    connection: NativeConnection {
      nativeClient: [External: 58c7630],
      referenceHolders: Set(0) {}
    },
    workflowsPath: '/home/val/Workspace/temporalio/samples-typescript/mutex/lib/workflows.js',
    activities: {
      useAPIThatCantBeCalledInParallel: [AsyncFunction: useAPIThatCantBeCalledInParallel]
    },
    taskQueue: 'test1/5/2023',
    shutdownGraceTimeMs: 10000,
    stickyQueueScheduleToStartTimeoutMs: 10000,
    isolateExecutionTimeoutMs: 5000,
    maxHeartbeatThrottleIntervalMs: 60000,
    defaultHeartbeatThrottleIntervalMs: 30000,
    loadedDataConverter: {
      payloadConverter: [DefaultPayloadConverter],
      failureConverter: [DefaultFailureConverter],
      payloadCodecs: []
    }
  }
}
2023-01-05T15:50:07.223Z [INFO] asset workflow-bundle-f8ba8cbc13e1ecef2505.js 703 KiB [emitted] [immutable] (name: main)
2023-01-05T15:50:07.224Z [INFO] modules by path ./node_modules/@temporalio/ 164 KiB
2023-01-05T15:50:07.224Z [INFO]   modules by path ./node_modules/@temporalio/common/lib/ 69 KiB
2023-01-05T15:50:07.224Z [INFO]     modules by path ./node_modules/@temporalio/common/lib/*.js 45 KiB 13 modules
2023-01-05T15:50:07.224Z [INFO]     modules by path ./node_modules/@temporalio/common/lib/converter/*.js 24 KiB 5 modules
2023-01-05T15:50:07.224Z [INFO]   modules by path ./node_modules/@temporalio/workflow/ 93.1 KiB
2023-01-05T15:50:07.224Z [INFO]     ./node_modules/@temporalio/workflow/lib/worker-interface.js 9.81 KiB [built] [code generated]
2023-01-05T15:50:07.224Z [INFO]     + 12 modules
2023-01-05T15:50:07.224Z [INFO]   ./node_modules/@temporalio/worker/lib/workflow-log-interceptor.js 2.27 KiB [built] [code generated]
2023-01-05T15:50:07.224Z [INFO] modules by path ./lib/ 3.44 KiB
2023-01-05T15:50:07.224Z [INFO]   ./lib/workflows-autogenerated-entrypoint.cjs 535 bytes [built] [code generated]
2023-01-05T15:50:07.224Z [INFO]   ./lib/workflows.js 2.92 KiB [built] [code generated]
2023-01-05T15:50:07.224Z [INFO] __temporal_custom_payload_converter (ignored) 15 bytes [built] [code generated]
2023-01-05T15:50:07.224Z [INFO] __temporal_custom_failure_converter (ignored) 15 bytes [built] [code generated]
2023-01-05T15:50:07.224Z [INFO] ./node_modules/long/umd/index.js 43.1 KiB [built] [code generated]
2023-01-05T15:50:07.224Z [INFO] ./node_modules/ms/index.js 2.95 KiB [built] [code generated]
2023-01-05T15:50:07.224Z [INFO] webpack 5.75.0 compiled successfully in 412 ms
2023-01-05T15:50:07.226Z [INFO] Workflow bundle created { size: '0.69MB' }
2023-01-05T15:50:07.459909Z  INFO temporal_sdk_core::worker: Initializing worker task_queue=test1/5/2023 namespace=default
2023-01-05T15:50:07.460Z [INFO] Worker state changed { state: 'RUNNING' }

```

So I'd prefer to keep this. But happy to remove it if we want to remove this output across the board.

mutex/src/start-lock-workflow.ts
vkarpov15: @bergundy I wasn't able to figure out how to handle signalWithStart from an activity like the Go SDK examples do. It doesn't look like TS SDK's `Context` class lets you access the Client. And there doesn't seem to be any examples of that to work off of in the SDK tests or sample apps.

I'm thinking a cleaner approach would be to just start the lock Workflow from the same client script as the test Workflow like I did here: https://github.com/temporalio/samples-typescript/pull/222/commits/f8102f027823d45540dd0f86dcc141647f329cd3 . Are you ok with that or is there an alternative approach you prefer?

lorensr: Ah I see what Go does

https://github.com/temporalio/samples-go/blob/e7f07468217806f57b22205692c4d70608b3e68b/mutex/mutex_workflow.go#L138

We tell people to import from `@temporalio/client` and create a new Client. Seems nice to add it to Context though? @bergundy

bergundy: I think we'll end up doing that, Spencer and I have both mentioned it before.
It's useful.
In the meantime, you can inject a client via interceptors or though a class constructor.
Here's an interceptor we use for testing the SDK: https://github.com/temporalio/sdk-typescript/blob/main/packages/test/src/activities/interceptors.ts

mutex/src/test/workflows.test.ts
bergundy: I'm not religious about this, up to you.

mutex/src/start-lock-workflow.ts
lorensr: For this sample, to me interceptors feels like an unnecessary complication, and creating a Client and injecting it like below is simpler

https://github.com/temporalio/samples-typescript/blob/75cb3264bf28517a4aeff95edcbac147faba1c2d/activities-dependency-injection/src/worker.ts#L17

bergundy: I generally tend to agree.
If you want you can use a class instead of this factory pattern.

mutex/src/test/workflows.test.ts
vkarpov15: Fixed in https://github.com/temporalio/samples-typescript/pull/222/commits/86d3b3212a58c408e4d55919f18c0051b690eae8, please check it out :+1:

vkarpov15:
@bergundy @lorensr please take a look at latest commit, I made almost all requested changes.

mutex/src/workflows.ts
vkarpov15: Do you have an example of what this would look like?

lorensr: if requests.length
  continueAsNew

vkarpov15:
Got ok from @bergundy in slack

PR #211, opened 2022-11-19
Add food delivery sample app

lorensr:
image

image

image

PR #225, opened 2023-01-10
run test and lint as a separate workflow per sample

vkarpov15:
## What was changed

Instead of one workflow that tests all samples and one workflow that lints all samples, create a separate GitHub workflow for linting and testing each individual sample

## Why?

Loren and I discussed this approach on Slack primarily because of the out of disk space errors we ran into on #211, #220, and #169. In addition to avoiding the risk of running out of disk space, this approach should be faster and more durable.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


lorensr (approved):
image

PR #228, opened 2023-01-13
working example for temporalio/documentation#1860

vkarpov15:
## What was changed

Not sure if we should commit this as a sample or just add it to dev guide as originally suggested in temporalio/documentation#1860. But here's a working example of using the same logger import in both activity and workflow based on @lorensr 's suggestions. Here's what the output looks like:

image

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


lorensr (approved):
I like it—I'd just say in README how it's an example of a shared helper function, and how it's different from `instrumentation`.

PR #233, opened 2023-01-20
proof of concept for separate prettier config for each sample

### LINKED ISSUE #231: [Feature Request] Add prettier to samples

Currently we have prettierrc and prettierignore at root and run prettier at root. 

- It's the longest CI step (9m vs 1m for others)
- It doesn't end up in repos created with `@temporalio/create`

- Add config files to each repo (maybe a shared prettierrc and individual prettierignore)
- Rm prettier build step and add it to individual-lint

vkarpov15:
Re: #231

## What was changed

Shared .prettierrc.json and .prettierignore for every sample. prettier as a devDependency in every sample. Run `npx prettier . --check` as part of lint-individual workflow. I needed to also upgrade prettier in the top-level package.json, because of potential mismatch between top-level prettier and package-level prettier (see https://github.com/temporalio/samples-typescript/actions/runs/3971200857/jobs/6807821215)

We need a .prettierignore for every sample, because otherwise prettier will always throw on compiled TS code `lib/`.

## Checklist

1. Closes #231

2. How was this tested:

3. Any docs updates needed?


lorensr (approved):
any benefit to being JSON vs `.prettierrc`? I like the shorter filename and contents of latter

vkarpov15:
Nope, I switched to using `.prettierrc` instead of `.json` in ed1b491

vkarpov15:
@lorensr can you please remove nestjs-counter test-individual from expected checks? That's blocking merging this PR:

image

lorensr:
Hah 😄 thanks!

PR #239, opened 2023-01-26
Fix lint; prettify

lorensr:
Also testing whether `npm ci` in CI is necessary


vkarpov15 (approved):
Nice :+1: looks like this makes tests a good bit faster

Issue #240, opened 2023-01-26
[Bug] unable to run food-delivery example


flerro:
### What are you really trying to do?

Start the example in the `food-delivery` folder using instructions provided in the related `README.md` file.

### Describe the bug

The `npm run dev` command in the `food-delivery` folder display some errors.

This an excerpt of the `npm run dev` command output:

```
worker:dev: /Users/flerro/Labs/temporal/samples-typescript/food-delivery/node_modules/ts-node/src/index.ts:859
worker:dev:     return new TSError(diagnosticText, diagnosticCodes, diagnostics);
worker:dev:            ^
worker:dev: TSError: ⨯ Unable to compile TypeScript:
worker:dev: src/worker.ts(4,49): error TS2307: Cannot find module 'common/lib/temporal-connection' or its corresponding type declarations.
worker:dev:
worker:dev:     at createTSError (/Users/flerro/Labs/temporal/samples-typescript/food-delivery/node_modules/ts-node/src/index.ts:859:12)
worker:dev:     at reportTSError (/Users/flerro/Labs/temporal/samples-typescript/food-delivery/node_modules/ts-node/src/index.ts:863:19)
worker:dev:     at getOutput (/Users/flerro/Labs/temporal/samples-typescript/food-delivery/node_modules/ts-node/src/index.ts:1077:36)
worker:dev:     at Object.compile (/Users/flerro/Labs/temporal/samples-typescript/food-delivery/node_modules/ts-node/src/index.ts:1433:41)
worker:dev:     at Module.m._compile (/Users/flerro/Labs/temporal/samples-typescript/food-delivery/node_modules/ts-node/src/index.ts:1617:30)
worker:dev:     at Module._extensions..js (node:internal/modules/cjs/loader:1293:10)
worker:dev:     at Object.require.extensions.<computed> [as .ts] (/Users/flerro/Labs/temporal/samples-typescript/food-delivery/node_modules/ts-node/src/index.ts:1621:12)
worker:dev:     at Module.load (node:internal/modules/cjs/loader:1096:32)
worker:dev:     at Function.Module._load (node:internal/modules/cjs/loader:935:12)
worker:dev:     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:84:12) {
worker:dev:   diagnosticCodes: [ 2307 ]

### Minimal Reproduction

1. Clone the repository 
2. Move into the `food-delivery` directory
3. Install the dependencies via `npm`
[...]

PR #241, opened 2023-01-26
fix(food-delivery): unable to run the example

flerro:
In the `worker.ts` for the food-delivery example, an invalid path in `import` prevents applications from starting correctly

## What was changed

An `import` in `worker.ts` file for the food-delivery example

## Why?

The `import` directive referenced an invalid path

## Checklist

1. Closes [#240](https://github.com/temporalio/samples-typescript/issues/240) 

2. How was this tested: start the applications via `npm run dev`

### DIFF
--- a/food-delivery/apps/worker/src/worker.ts
+++ b/food-delivery/apps/worker/src/worker.ts
@@ -1,7 +1,7 @@
 import { NativeConnection, Worker } from '@temporalio/worker'
 import * as activities from 'activities'
 import { taskQueue } from 'common'
-import { namespace, getConnectionOptions } from 'common/lib/temporal-connection'
+import { namespace, getConnectionOptions } from '../../../packages/common/temporal-connection'
 
 async function run() {
   const connection = await NativeConnection.connect(getConnectionOptions())


lorensr:
`'common/lib/temporal-connection'` is the recommended way in turborepo to require other packages in the monorepo. If you build before dev, it works (lib/ is created during build). Looks like I included the build step in .post-create but not the README.

It would be better if we told turborepo what it needed to build before it runs dev.

Maybe something like this in `turbo.json`, but it's not working for me:

```
    "dev": {
      "dependsOn": ["^build"]
    },
```

https://turbo.build/repo/docs/core-concepts/monorepos/running-tasks

vkarpov15:
@lorensr I tried it out and I can confirm `"dependsOn": ["^build"],` doesn't work. But `"dependsOn": ["build"],` **does** work. Not sure what the up-carat syntax is. But [here's some docs on running tasks before 'dev' in Turbo](https://turbo.build/repo/docs/handbook/dev#running-tasks-before-dev).

image

lorensr:
😄 ^build is supposed to only build deps, not all packages. but that's fine, build works for me:  #243

flerro:
Thanks @vkarpov15 and @lorensr a lot for looking into it! 

Since the solution proposed in #243 is the best one, I will close this PR

lorensr:
Thanks for pointing it out, and thanks @vkarpov15 for fixing! ☺️

PR #245, opened 2023-01-31
chore: Update Temporal SDK dependencies to 1.6.0

mjameswh:
## What changed

- Updated all @temporalio dependencies to ^1.6.0

## Notes

- I deleted some package-lock.json files that were still referencing @grpc/grpc-js 1.8.0 after update.


mjameswh:
@vkarpov15 @lorensr What'd the status on commiting package-lock.json files in samples? I see that at some point, Val did commit all files, but they, are are mostly missing. Why?

If we still want to commit these files, do we have a procedure to regenerate them?

bergundy:
We can't commit package-lock files since swc installs platform specific modules resulting in broken installations.

bergundy (approved):

PR #250, opened 2023-02-06
Fix hello world

lorensr:


kimschles (approved):
I tested these changes, and they work as expected.

PR #266, opened 2023-05-12
Add vscode-debugger sample

lorensr:


README.md
```diff
@@ -126,7 +126,9 @@ and you'll be given the list of sample options.
 #### Production APIs
 
 - [**Production Build**](./production): Build code in advance for faster Worker startup times.
-- [**Debugging**](https://docs.temporal.io/application-development/debugging#debug): The [replay-history](./replay-history) sample shows how to retrieve Workflow Event History and debug it using the `runReplayHistory` Worker API ([video](https://youtu.be/fN5bIL7wc5M)).
+- [**Debugging**](https://docs.temporal.io/application-development/debugging#debug):
+  - The [replay-history](./replay-history) sample shows how to retrieve Workflow Event History and debug it using the `runReplayHistory` Worker API ([video](https://youtu.be/fN5bIL7wc5M)).
```
bergundy, line 130: I would just remove it at this point and (if there's a reference to it from YT) note in the video description that the sample has been replaced by this newer and shinier one.

vscode-debugger/README.md
```diff
@@ -0,0 +1,12 @@
+# VS Code Debugger
+
+Use our [VS Code plugin](https://marketplace.visualstudio.com/items?itemName=temporal-technologies.temporalio) to debug a running or completed Workflow Execution. See this blog post for a walkthrough:
```
bergundy, line 3: Wanna record a video too?

bergundy: Can just be screen capture, but I guess that might go on the extension README.

vscode-debugger/src/client.ts
```diff
@@ -0,0 +1,23 @@
+import { Client } from '@temporalio/client';
+import { nanoid } from 'nanoid';
+import wait from 'waait';
```
bergundy, line 3: If we don't care about node 14 you can use
```
import { setTimeout } from 'node:timers/promises';
```

vscode-debugger/src/client.ts
```diff
@@ -0,0 +1,23 @@
+import { Client } from '@temporalio/client';
+import { nanoid } from 'nanoid';
```
bergundy, line 2: And we can just use `crypto.randomUUID()` to remove this dependency.

bergundy (approved):
Thanks! ❤️

vscode-debugger/README.md
lorensr: Will link video from post

vscode-debugger/src/client.ts
lorensr: i like keeping 14 support in samples as long as SDK supports it

PR #268, opened 2023-05-29
Rename and clarify sticky activities sample

lorensr:


lorensr:
The link to https://docs.temporal.io/tasks/#sticky-execution was inaccurate. Will also update the [python sample](https://github.com/temporalio/samples-python/tree/main/activity_sticky_queues) once this is reviewed/merged

PR #271, opened 2023-06-25
fix monorepos-folders link

msjaber:
## What was changed
- fix monorepos-folders link


lorensr:
Thanks for the fix! 😊

PR #276, opened 2023-07-05
Set Temporal CLI address env variable

feedmeapples:
## What was changed

Set Temporal CLI address env variable
Upgrade UI

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,6 +59,7 @@ services:
       - temporal
     environment:
       - TEMPORAL_CLI_ADDRESS=temporal:7233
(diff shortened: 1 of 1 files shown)


mjameswh:
Thanks!

May I ask however why you care about that docker-compose file? I had completely forgot we have it, but I would consider removing it entirely, unless we still have something that wouldn't work using CLI dev server.

lorensr:
Think used for one/both of these? Agree them using CLI would be better

image

PR #279, opened 2023-07-08
chore: Update to TS SDK 1.8.1

mjameswh:
## What was changed

TBD


activities-cancellation-heartbeating/src/activities.ts
```diff
@@ -2,20 +2,21 @@
 import { CancelledFailure, Context } from '@temporalio/activity';
 
 export async function fakeProgress(sleepIntervalMs = 1000): Promise<void> {
+  const { log, info, sleep, heartbeat } = Context.current();
   try {
     // allow for resuming from heartbeat
-    const startingPoint = Context.current().info.heartbeatDetails || 1;
-    console.log('Starting activity at progress:', startingPoint);
+    const startingPoint = info.heartbeatDetails || 1;
+    log.info(`Starting activity at progress: ${startingPoint}`);
```
bergundy, line 9: Better use structured logs
```suggestion
    log.info('Starting activity at progress', { startingPoint });
```

activities-cancellation-heartbeating/src/activities.ts
```diff
@@ -2,20 +2,21 @@
...
     for (let progress = startingPoint; progress <= 100; ++progress) {
       // simple utility to sleep in activity for given interval or throw if Activity is cancelled
       // don't confuse with Workflow.sleep which is only used in Workflow functions!
-      console.log('Progress:', progress);
-      await Context.current().sleep(sleepIntervalMs);
-      Context.current().heartbeat(progress);
+      log.info(`Progress: ${progress}`);
```
bergundy, line 13:
```suggestion
      log.info('Progress', { progress });
```

activities-cancellation-heartbeating/src/activities.ts
```diff
@@ -2,20 +2,21 @@
...
+      await sleep(sleepIntervalMs);
+      heartbeat(progress);
     }
   } catch (err) {
     if (err instanceof CancelledFailure) {
-      console.log('Fake progress activity cancelled', err.message);
+      log.warn(`Fake progress activity cancelled: ${err.message}`);
```
bergundy, line 19: Same here please

custom-logger/README.md
```diff
@@ -0,0 +1,187 @@
...
+1. `npm run start.watch` to start the Worker.
+1. In another shell, `npm run workflow` to run the Workflow.
+
+<details>
+<summary>
+Sample worker output
+</summary>
+
+```
+2023-09-05T18:19:39.646Z [undefined] info: Creating worker {
```
bergundy, line 21: undefined?

custom-logger/src/worker.ts
```diff
@@ -0,0 +1,54 @@
...
+        ...entry.meta,
+      });
+    }),
+    // Telemetry options control how logs are exported out of Rust Core.
+    telemetryOptions: {
+      logging: {
+        // By default, Core logs go directly to console. Setting the `forward` property here (to an
+        // empty object) enables forwarding of logs from Rust Core to the Node.js logger.
+        forward: {},
+        // This filter determines which logs should be forwarded from Rust Core to the Node.js logger.
```
bergundy, line 30: I would add a note that you would likely not want this to be set to DEBUG in a production setting. Probably WARN is enough (and in the line above).

bergundy (approved):

activities-cancellation-heartbeating/src/activities.ts
mjameswh: Good point. Will fix all of these.

custom-logger/README.md
mjameswh: Good catch. The original sample was adding a `label` metadata (`worker`, `workflow` or `activity`). That required registering different loggers on each context, which I definitely want to avoid.

For the moment, I fixed this with an heuristic in the logger adaptor ([see here](https://github.com/temporalio/samples-typescript/pull/279/files#diff-ddf423b693396f737a7ca3e8f5e525a9fb5e213493dd1530bfa853bcbfa31abfR18)), just so that we can move on to merging this PR, but I can easily imagine cases where that would guess incorrectly.

I think we should provide something similar to that 'label' property out of the box.

PR #285, opened 2023-08-11
Improve hello-world sample

bergundy:
## What was changed

Restructured the hello-world sample to make it more explicit

## Why?

I noticed some users struggle to with the implicitness in the samples and have a hard time discovering some of the more advanced configuration knobs.

### DIFF
--- a/hello-world/src/worker.ts
+++ b/hello-world/src/worker.ts
@@ -4,6 +4,9 @@ import * as activities from './activities';
 
 async function run() {
   // Step 1: Establish a connection with Temporal server.
+  //
+  // Worker code uses `@temporalio/worker.NativeConnection`.
+  // (But in your application code it's `@temporalio/client.Connection`.)
   const connection = await NativeConnection.connect({
     address: 'localhost:7233',
     // TLS and gRPC metadata configuration goes here.
--- a/hello-world/src/worker.ts
+++ b/hello-world/src/worker.ts
@@ -1,22 +1,31 @@
 // @@@SNIPSTART typescript-hello-worker
-import { Worker } from '@temporalio/worker';
+import { NativeConnection, Worker } from '@temporalio/worker';
 import * as activities from './activities';
 
 async function run() {
-  // Step 1: Register Workflows and Activities with the Worker and connect to
-  // the Temporal server.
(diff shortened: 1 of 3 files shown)


Sushisource (approved):
Makes sense to me

hello-world/src/worker.ts
```diff
@@ -1,22 +1,31 @@
 // @@@SNIPSTART typescript-hello-worker
-import { Worker } from '@temporalio/worker';
+import { NativeConnection, Worker } from '@temporalio/worker';
 import * as activities from './activities';
 
 async function run() {
-  // Step 1: Register Workflows and Activities with the Worker and connect to
-  // the Temporal server.
+  // Step 1: Establish a connection with Temporal server.
```
bergundy, line 6:
```suggestion
  // Step 1: Establish a connection with Temporal server.
  //
  // NativeConnection is implemented in Rust and is used by Core SDK which powers the Worker.
```

dandavison: Is there any benefit in mentioning Rust and Core SDK here? It's strictly implementation detail, so if these changes are aimed at helping less expert users it may be best not to. E.g. it might trigger thoughts like "Do I need Rust?", "Where do I get Core SDK?"

mjameswh: Agreed it might be pertinent to make this text just a little bit easier to understand for new users, but we somehow need to make it clear that `Connection` is for client calls, and `NativeConnection` is for Workers. And unfortunately, the `Native` prefix here may sounds a bit mysterious and isn't helping remembering that difference, unless the user understands that the Worker relies on some native code (aka Core SDK, coded in Rust).

dandavison: I see. I'd suggest something like

```
Worker code uses `NativeConnection`. 
(But in your application code it's just `Connection`.)
```

bergundy: I like this

bergundy: I really hope we can deprecate this class soon when Core can delegate gRPC calls to lang.

bergundy:
@mjameswh looks like CI is unstable.
I will open an issue to update this for all samples.

PR #287, opened 2023-08-23
add SSE example

nubunto:
## What was changed

Added a very simple SSE sample.

## Why?

Seems like something people usually want to know how to work out.


mjameswh:
Great! Thanks @nubunto for contributing this sample.

I think we really need to demonstrate this concept is a setup that supports the possibly of having more than one HTTP server running at the same time. The main difficulty in that case is that the activity task to send a message to user/connection X need to be routed to the appropriate activity worker, ie. the one that lives in the process that own that HTTP connection.

Implementation wise, this is not very complex (per host task queue), but that requires that workflows sending messages knows the list of user/activity task queues to which they need post messages.

What I have in mind would be to have something like a "Chat Room Workflow". When establish an SSE connection, the user would have to tell the name of the chat room it want to join (eg. to make it easy, that could be a segment of the URL). The HTTP handler add the connection to its local pool of SSE connection, then do `signalWithStart` to the appropriate chat room workflow execution (ie. workflow id is derived from chat room name), and send its task queue name + some form of user identifier as argument on the subscribe signal.

What do you think?

nubunto:
that was on the back of my mind as I finished and tested the implementation, and I think it's very reasonable! I'll update the PR accordingly as soon as I have the time!

nubunto:
Another possible way of achieving the same end result of horizontally scaling the HTTP server is by leveraging something like Redis pub sub as a lightweight message broker. Usually this is not what I'd do if I'm going for durability, but since the message publishing is backed by Temporal it might be an relatively lightweight solution.

nubunto:
@mjameswh I implemented your suggestion! took me a while, sorry about that. Please, take a look whenever possible and let me know what you think!

mjameswh:
Great! Will try to review by the end of this week. Thanks for your work on this.

PR #297, opened 2023-10-08
feat: AWS SQS Example

### LINKED ISSUE #296: [Feature Request] AWS SQS Sample

Using Temporal IO alongside an AWS FIFO queue (or AWS queue in general) can be beneficial for several reasons:

- **Decoupling of Services**: In environments where two services cannot communicate over the same local network due to being in different clusters or policy constraints, a queue acts as a mediator. AWS SQS can store messages until they're consumed, ensuring that messages aren't lost even if the consuming service isn't immediately available.

- **Integration with Existing Architecture:** For organizations that have already invested in AWS and use SQS queues, integrating Temporal can enhance the processing capabilities without a complete overhaul. Temporal can be introduced to handle the business logic, retries, and workflows, while SQS continues to act as the message broker.

- **Ordered Processing:** FIFO queues ensure that messages are processed in the order they are sent, which is crucial in scenarios like financial transactions or data synchronization.

Sample folder that shows some basic setup!

michaelangeloio:
## What was changed
This PR introduces a sample showcasing the integration of Temporal with AWS SQS, specifically focusing on FIFO queues. 

## Why?
Using Temporal IO alongside an AWS FIFO queue (or AWS queue in general) can be beneficial for several reasons:

- **Decoupling of Services:**  In environments where two services cannot communicate over the same local network due to being in different clusters or policy constraints, a queue acts as a mediator. AWS SQS can store messages until they're consumed, ensuring that messages aren't lost even if the consuming service isn't immediately available.

- **Integration with Existing Architecture:** For organizations already invested in AWS and using SQS queues, integrating Temporal can enhance the processing capabilities without a complete overhaul. Temporal can be introduced to handle the business logic, retries, and workflows, while SQS continues to act as the message broker.

- **Ordered Processing:** FIFO queues ensure that messages are processed in the order they are sent, which is crucial in scenarios like financial transactions or data synchronization.

Also, customers will be happy to see it's possible to integrate temporal into their existing stack 😉 

## Checklist

1. Closes #296 

2. How was this tested:
Follow the steps for spinning up the stack, and everything should work smoothly (should you provide the correct AWS inputs)!

I can write some unit tests; however, I doubt this would be beneficial as this is simply a sample.

3. Any docs updates needed?
Included a README as part of this PR.

### DIFF
--- a/aws-sqs-node/src/utils.ts
+++ b/aws-sqs-node/src/utils.ts
@@ -1,0 +1,61 @@
+import { SQSClient } from '@aws-sdk/client-sqs'
+import 'dotenv/config'
+import pinoLogger, { Logger } from 'pino'
+import { z } from 'zod'
+
+const envSchema = z.object({
+  AWS_ACCESS_KEY_ID: z.string(),
+  AWS_SECRET_ACCESS_KEY: z.string(),
+  AWS_REGION: z.string(),
+  AWS_ACCOUNT_ID: z.string(),
+  QUEUE_NAME: z.string(),
+})
+
+let fetchedEnvs: z.infer<typeof envSchema> | null = null
+
+export const getEnv = () => {
+  if (!fetchedEnvs) {
+    const { AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION, AWS_ACCOUNT_ID, QUEUE_NAME } = process.env
+    const envs = {
+      AWS_ACCESS_KEY_ID,
+      AWS_SECRET_ACCESS_KEY,
+      AWS_REGION,
+      AWS_ACCOUNT_ID,
+      QUEUE_NAME,
+    }
+    const result = envSchema.safeParse(envs)
+    if (!result.success) {
+      throw new Error(result.error.message)
+    }
+    fetchedEnvs = result.data
+    return result.data
+  }
+  return fetchedEnvs
+}
+
+export const getQueueUrl = () => {
+  const env = getEnv()
+  return `https://sqs.${env.AWS_REGION}.amazonaws.com/${env.AWS_ACCOUNT_ID}/${env.QUEUE_NAME}`
+}
+
+let logger: Logger
+export const getLogger = () => {
+  if (!logger) {
+    logger = pinoLogger({
+      level: process.env.NODE_ENV === 'production' ? 'info' : 'debug',
+    })
+  }
+  return logger
+}
+
+export const getSQSClient = () => {
+  const env = getEnv()
+  const sqs = new SQSClient({
+    region: env.AWS_REGION,
+    ...(process.env.NODE_ENV === 'production' ?  {
+      accessKeyId: env.AWS_ACCESS_KEY_ID,
+      secretAccessKey: env.AWS_SECRET_ACCESS_KEY,
+    } : {}),
+  })
+  return sqs
+}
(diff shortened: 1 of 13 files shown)


aws-sqs-node/src/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!fetchedEnvs) {
+    const { AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION, AWS_ACCOUNT_ID, QUEUE_NAME } = process.env
+    const envs = {
+      AWS_ACCESS_KEY_ID,
+      AWS_SECRET_ACCESS_KEY,
+      AWS_REGION,
+      AWS_ACCOUNT_ID,
+      QUEUE_NAME,
+    }
+    const result = envSchema.safeParse(envs)
```
michaelangeloio, line 26: I included this logic so the average user gets an error if they have not configured their AWS env variables, which is needed to make this sample run.

mjameswh:
Thank you very much for this sample project. I'll review next week, but in the mean time, that looks like a very interesting example.

My only concern at this point is either we really want to maintain this as part of this samples repo. We've been recently discussing about restructuring the repo so that it is both easier to maintain and better fulfill its purpose. But let me check.

aws-sqs-node/src/worker.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    for (const message of data.Messages) {
+      const { nanoid } = await import('nanoid')
+      // keep in mind, you will have N temporal connections here, 
+      // one for each poller
+      const client = await getTemporalClient()
+      logger.info({ message }, 'Starting workflow')
+      await client.workflow.start(helloWorld, {
+        args: [message],
+        workflowId: nanoid(),
```
lorensr, line 48: Can you deterministically generate a workflowId from the message data, so that when there are duplicate messages, multiple workflows aren't created?

aws-sqs-node/src/worker.ts
```diff
@@ -0,0 +1,111 @@
...
+        taskQueue: 'hello-world-queue',
```
lorensr, line 49:
```suggestion
        // If a duplicate event might arrive after the Workflow started by the original event has completed, 
        // set this reuse policy to prevent duplicate Workflows:
        // workflowIdReusePolicy: WorkflowIdReusePolicy .WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE,
        taskQueue: 'hello-world-queue',
```

aws-sqs-node/src/worker.ts
michaelangeloio: Yes, you can! This example is FIFO based though, which typically is already de-duplicated. We can try using the message Id that comes from this data

nubunto:
@michaelangeloio amazing example, thanks! Been wanting to implement something similar lately.

What are your thoughts on using `worker_threads` instead of `cluster`?

michaelangeloio:
> @michaelangeloio amazing example, thanks! Been wanting to implement something similar lately.
> 
> What are your thoughts on using `worker_threads` instead of `cluster`?

Can you provide some context? You can theoretically use worker_threads, however, drawbacks include having to pass a file (that's compiled from TS to js) or string into the worker thread.

nubunto:
I'm about to port an existing SQS worker to our already existing Temporal infrastructure. I had a couple strategies in my mind:

1. fire "poller workflows", that would basically poll the queue using a long-running activity, process messages, and call `continueAsNew` periodically. Pros: I only use a single worker. Cons: I have workflows that run forever, which are a bit harder to update than normal.
2. have my current SQS worker read messages, and fire a workflow for the message. The workflow then deletes the message once it's done processing. Pros: easier to implement. Cons: two separate services instead of one.
3. your approach, which is kind of a mashup of the previous two. But my first instinct was to open one thread per poller, instead of a process. However, you have a point, cluster might be a good fit here.

PR #305, opened 2023-11-16
Empty project sample - contains no Workflow or Activity definitions

napcs:
An empty project so we can generate a blank project with all the configuration in place, but no Workflows and Activities.


empty/.post-create
bergundy, line 1: I would add some more guidance to this post create message

bergundy (approved):
LGTM, had a small suggestion.

empty/.post-create
napcs: Added.

empty/.post-create
```diff
@@ -0,0 +1,25 @@
+To begin development, install the Temporal CLI:
+
+  Mac: {cyan brew install temporal}
+  Other: Download and extract the latest release from https://github.com/temporalio/cli/releases/latest
+
+Start Temporal Server:
+
+  {cyan temporal server start-dev}
```
angelazhou32, line 8: I would recommend adding `--db-filename clusterdata.db` to this command or at least mentioning that this is how we would get Workflows to persist so in the case if the cluster crashed, or if the user shut it down before a break and started it up later, all of their Workflow Executions (and any other data) would be maintained

empty/.post-create
```diff
@@ -0,0 +1,25 @@
...
+
+  Mac: {cyan brew install node@16}
+  Other: https://nodejs.org/en/download/
+
+Then, in the project directory:
+
+* Add your Activity Definitions to `src/activities.ts`.
+* Add your Workflow Definitions to `src/workflows.ts`.
+* Modify the `src/client.ts` file and replace `YOUR_WORKFLOW` with the name of your Workflow.
+* Optionally, add activity and tests to the `src/mocha` directory in files with the extension `.test.ts`.
```
angelazhou32, line 20: Maybe "Add Activity and Workflow tests"? This statement you have makes it sound like to add Activities to that directory

empty/src/client.ts
```diff
@@ -0,0 +1,40 @@
...
+    taskQueue: TASK_QUEUE_NAME,
+    // type inference works! args: [name: string]
+    args: ['Temporal'],
+    // in practice, use a meaningful business ID, like customerId or transactionId
+    workflowId: 'workflow-' + nanoid(),
+  });
+  console.log(`Started workflow ${handle.workflowId}`);
+
+  // optional: wait for client result
+  console.log(await handle.result()); // Hello, Temporal!
```
angelazhou32, line 33: Maybe remove 'Hello, Temporal!' because we don't know that the result of the Workflow is going to be.

empty/README.md
```diff
@@ -0,0 +1,22 @@
...
+* Add your Activity Definitions to `src/activities.ts`.
+* Add your Workflow Definitions to `src/workflows.ts`.
+* Modify the `src/client.ts` file and replace `YOUR_WORKFLOW` with the name of your Workflow.
+* Add activity and tests to the `src/mocha` directory in files with the extension `.test.ts`.
+
+## Running the code
+
+Install dependencies with `npm install`.
+
+Run `temporal server start-dev` to start [Temporal Server](https://github.com/temporalio/cli/#installation).
```
angelazhou32, line 14: Same as my comment above with specifying the DB filename, port location etc.

empty/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+The `package.json` file contains scripts for running the client, the Worker, and tests.
+
+1. In a shell, run `npm run start.watch` to start the Worker and reload it when code changes..
```
angelazhou32, line 18: Remove '..'

empty/src/worker.ts
```diff
@@ -0,0 +1,40 @@
...
+  const worker = await Worker.create({
+    connection,
+    namespace: 'default',
+    taskQueue: TASK_QUEUE_NAME,
+    // Workflows are registered using a path as they run in a separate JS context.
+    workflowsPath: require.resolve('./workflows'),
+    activities,
+  });
+
+  // Step 3: Start accepting tasks on the `hello-world` queue
```
angelazhou32, line 25: 'your-task-queue-name' task queue

Issue #310, opened 2023-11-19
Make the worker specific task queues work (this.cancelled undefined)


giuliohome:
Please [refer to my comment here](https://github.com/temporalio/samples-typescript/issues/309#issuecomment-1817974209).

https://github.com/temporalio/samples-typescript/blob/98853010a11fc3c8a3dda49c7f0cdaa0e134fb32/worker-specific-task-queues/src/activities.ts#L17

giuliohome:
It does work when I transform it into all sync without async activities (without importing sleep etc... see below)


```ts
export function createActivitiesForSameWorker() {
  return {
    async downloadFileToWorkerFileSystem(url: string, path: string): Promise<void> {
      const { log } = Context.current();
      log.info('Downloading and saving', { url, path });
      // Here's where the real download code goes
      const body = 'downloaded body';
      // await sleep(3000);
      syncfs.writeFileSync(path, body);
    },
    async workOnFileInWorkerFileSystem(path: string): Promise<void> {
      const { log } = Context.current();
      const content = await fs.readFile(path);
      const checksum = createHash('md5').update(content).digest('hex');
      // await sleep(3000);
      log.info('Did some work', { path, checksum });
    },
    async cleanupFileFromWorkerFileSystem(path: string): Promise<void> {
      const { log } = Context.current();
      // await sleep(3000);
      log.info('Cleaning up temp file', { path });
      await fs.rm(path);
    },
  };
}
```

giuliohome:
Using the following delay instead of the above sleep solves the issue

```
function delay(ms: number) {
    return new Promise( resolve => setTimeout(resolve, ms) );
}
```

giuliohome:
Will be solved with next release of sdk-typescript by making sleep an arrow function.

mjameswh:
Resolved in [TS SDK 1.9.0](https://github.com/temporalio/sdk-typescript/releases/tag/v1.9.0).

PR #311, opened 2023-11-19
Fix issue #310

giuliohome:
## What was changed
Using delay instead of sleep in Worker-Specific Task Queues

## Why?
See the related issue https://github.com/temporalio/samples-typescript/issues/310

## Checklist
Verify why importing sleep from current context is wrong.

1. Closes 310

2. How was this tested:
Running in Gitpod (it's already a one click configuration, nothing special to do)

3. Any docs updates needed?
It's only a one liner fix, no docs to be updated, really.

### DIFF
--- a/worker-specific-task-queues/src/activities.ts
+++ b/worker-specific-task-queues/src/activities.ts
@@ -10,17 +10,7 @@ export function createNormalActivities(uniqueWorkerTaskQueue: string) {
     },
   };
 }
-//  Workflow.sleep(...) may only be used from a Workflow Execution
-function delay(ms: number) {
-  return new Promise( resolve => setTimeout(resolve, ms) );
-}
-function tsleep(ms: number, cancelled: Promise<never>){
-  let handle: NodeJS.Timeout;
-  const timer = new Promise<void>((resolve) => {
-    handle = setTimeout(resolve, ms);
-  });
-  return Promise.race([cancelled.finally(() => clearTimeout(handle)), timer]);
-}
+
 export function createActivitiesForSameWorker() {
   return {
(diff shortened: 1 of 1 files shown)


giuliohome:
Explanation from ChatGPT 4:

Your observation is correct. The issue you're encountering is related to the context of this in JavaScript.

In the first example:

```ts
const { sleep } = Context.current();
// ...
await sleep(3000);
```
Here, you're destructuring sleep from the Context.current() object. When you do this, sleep is now a standalone function, and it loses its context (this). If sleep function uses this internally (which it does in your case, this.cancelled), it will be undefined because this no longer refers to the original Context.current() object.

In the second example:

```ts
const context = Context.current();
await context.sleep(3000);
```
Here, you're not destructuring sleep. Instead, you're calling sleep as a method on the context object. In this case, this inside sleep will correctly refer to the context object, and this.cancelled will be defined.

This is a common pitfall in JavaScript when dealing with context (this). It's always important to ensure that context is preserved when it's needed.

giuliohome:
Solved in main sdk-typescript by making those as arrow functions

https://github.com/temporalio/sdk-typescript/pull/1249

Waiting for next release of sdk-typescript

giuliohome:
I'll keep this open until the SDK-Typescript arrow function version of `sleep` is published. In the meantime, if the new release is not available soon, the sample can be fixed.

bergundy (commented):
Thanks!
This will be fixed in the next SDK release but if you want to merge this I'd prefer storing the context in a variable and using `ctx.sleep(...)` instead of defining `delay` and `tsleep`.

mjameswh:
The issue has been resolved in [TS SDK 1.9.0](https://github.com/temporalio/sdk-typescript/releases/tag/v1.9.0).

PR #318, opened 2023-11-27
fix typos for empty project instructions and code

napcs:
Unfortunately a handful of typos made their way into the empty project. @angelazhou32 is an excellent proofreader.

### DIFF
--- a/empty/README.md
+++ b/empty/README.md
@@ -4,8 +4,9 @@ This is an empty project that is scaffolded out when you run `npx @temporalio/cr
 
 * Add your Activity Definitions to `src/activities.ts`.
 * Add your Workflow Definitions to `src/workflows.ts`.
+* Set your task queue name in `src/shared.ts`.
 * Modify the `src/client.ts` file and replace `YOUR_WORKFLOW` with the name of your Workflow.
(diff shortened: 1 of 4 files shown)


empty/README.md
```diff
@@ -15,7 +16,7 @@ Run `temporal server start-dev` to start [Temporal Server](https://github.com/te
 
 The `package.json` file contains scripts for running the client, the Worker, and tests.
 
-1. In a shell, run `npm run start.watch` to start the Worker and reload it when code changes..
+1. In a shell, run `npm run start.watch` to start the Worker and reload it when code changes.
```
angelazhou32, line 19: Do we need to mention 'reload it when code changes' when we have `start.watch`? Because of the .watch, it should automatically detect code changes.

angelazhou32 (approved):

empty/README.md
napcs: The sentence may be poorly constructed. But it's saying that the command "starts the worker and reloads it".

angelazhou32: Oh gotcha, I thought that you meant to tell the user to reload that command every time there is a change

PR #328, opened 2024-01-16
Update to TS SDK 1.9

mjameswh:
## What was changed

- Updated all @temporalio/* dependencies to ^1.9.0
- Use Activity exported functions/objects instead of `Context.current().something()`
- Use `TestWorkflowEnvironment.createLocal` instead of `createTimeSkipping` where possible
- Remove usage of `defaultSinks`


.scripts/list-of-samples.json
```diff
@@ -9,12 +9,14 @@
...
     "dsl-interpreter",
     "ejson",
+    "empty",
     "encryption",
     "expense",
     "fetch-esm",
     "food-delivery",
     "grpc-calls",
     "hello-world",
+    "hello-world-esm",
```
lorensr, line 19: remove this and the dir.

also, you also needed this change:

https://github.com/temporalio/samples-typescript/pull/327/files#diff-88397926b71aad9ca766a35d4c240f95914117d55d5521d57ebe2ad7b0a390c2

before committing (the commit hook wrote/overwrote some files it shouldn't have in this PR).

lorensr (approved):
The current build error should be fixed by merging main

PR #333, opened 2024-01-26
Typo: unexepected -> unexpected

nurkiewicz:


lorensr:
Thanks for the fix! 😊

PR #334, opened 2024-01-26
[DRAFT] Add context propagation + custom log attributes sample

mjameswh:
## What was changed

- Add a sample that demonstrates using interceptors to propagate contextual data from the client, to workflows, to child workflows, and to activities. That sample also demonstrates how to use interceptors to add custom log attributes to log messages emitted using workflow.log and activity.log.

## Why?

- We no longer had any samples of how to develop TS interceptors.

- Context Propagation is a recurring support request.

- This sample may provide some ground to define built-in support for context propagation (see temporalio/sdk-typescript#998).


context-propagation/README.md
```diff
@@ -0,0 +1,175 @@
...
+1. `npm run start.watch` to start the Worker.
+1. In another shell, `npm run workflow` to run the Workflow.
+
+<details>
+<summary>
+Sample worker output
+</summary>
+
+```
+2023-09-05T18:19:39.646Z [worker] info: Creating worker {
```
mjameswh, line 26: Remove this output, not pertinent

PR #335, opened 2024-01-30
Update sample

dandavison:
Add an `executeUpdate` example.

### DIFF
--- a/update/src/client.ts
+++ b/update/src/client.ts
@@ -14,9 +14,11 @@ async function run() {
   });
   console.log(`Started workflow ${handle.workflowId}`);
 
+  // Send the 'fetchAndAdd' update with the value `1` as input.
   const prevValue = await handle.executeUpdate(fetchAndAdd, { args: [1] });
   console.log(`incrementing counter; previous value was ${prevValue}`);
 
+  // Send the 'fetchAndAdd' update with invalid input.
   const invalidArg = -1;
   try {
     await handle.executeUpdate(fetchAndAdd, { args: [invalidArg] });
@@ -24,8 +26,8 @@ async function run() {
     console.log(`Update argument ${invalidArg} was rejected: ${error}`);
   }
 
-  // Send the signal to allow the workflow to complete.
-  await handle.signal(done);
+  // Send the 'done' update to allow the workflow to complete.
+  await handle.executeUpdate(done);
 
   console.log(`Final counter value is ${await handle.result()}`);
 }
--- a/update/src/client.ts
+++ b/update/src/client.ts
@@ -1,0 +1,37 @@
+// @@@SNIPSTART typescript-update-client
+import { Connection, Client } from '@temporalio/client';
+import { counter, fetchAndAdd, done } from './workflows';
+import { nanoid } from 'nanoid';
+
+async function run() {
+  const connection = await Connection.connect({ address: 'localhost:7233' });
+  const client = new Client({ connection });
+
+  const handle = await client.workflow.start(counter, {
+    taskQueue: 'my-task-queue',
+    args: [],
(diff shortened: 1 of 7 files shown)


update/src/client.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const invalidArg = -1;
+  try {
+    await handle.executeUpdate(fetchAndAdd, { args: [invalidArg] });
+  } catch (error) {
+    console.log(`Update argument ${invalidArg} was rejected: ${error}`);
+  }
+
+  // Send the signal to allow the workflow to complete.
+  await handle.signal(done);
```
lorensr, line 28: maybe use another update?

update/src/workflows.ts
```diff
@@ -0,0 +1,31 @@
+// @@@SNIPSTART typescript-update-workflow
+import * as wf from '@temporalio/workflow';
+
+export const fetchAndAdd = wf.defineUpdate<number, [number]>('fetchAndAdd');
+export const done = wf.defineSignal('done');
+
+export async function counter(): Promise<number> {
+  let count = 0;
+  let shouldExit = false;
```
lorensr, line 9:
```suggestion
  let shouldComplete = false;
```

lorensr (approved):
Yay!

dandavison:
Thanks @lorensr. I made both your changes, and added some comments since the purpose of these is to be instructive.

update/src/client.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const handle = await client.workflow.start(counter, {
+    taskQueue: 'my-task-queue',
+    args: [],
+    workflowId: 'workflow-' + nanoid(),
+  });
+  console.log(`Started workflow ${handle.workflowId}`);
+
+  // Send the 'fetchAndAdd' update with the value `1` as input.
+  const prevValue = await handle.executeUpdate(fetchAndAdd, { args: [1] });
```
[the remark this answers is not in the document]
dandavison: Yes this sample is just for `executeUpdate`. I was thinking we'd add a second sample for `startUpdate`.

update/src/worker.ts
```diff
@@ -0,0 +1,21 @@
...
+import { NativeConnection, Worker } from '@temporalio/worker';
+
+async function run() {
+  const connection = await NativeConnection.connect({
+    address: 'localhost:7233',
+  });
+  const worker = await Worker.create({
+    connection,
+    namespace: 'default',
+    taskQueue: 'my-task-queue',
```
bergundy, line 11: Use a task queue that's unique for this sample. In case users run multiple samples, you want their workers to not collide.

update/src/workflows.ts
```diff
@@ -0,0 +1,43 @@
...
+  // Define an update validator that rejects negative inputs to the update.
+  const validator = (arg: number) => {
+    if (arg < 0) {
+      throw new Error('Argument must not be negative');
+    }
+  };
+
+  // Define the update handler implenting the fetchAndAdd operation. Note that
+  // an update can mutate workflow state, and return a value.
+  const handler = (arg: number) => {
```
bergundy, line 22: Worth also should that update handlers can be async and can sleep or run activities.

bergundy (commented):
Apologies for the late review.

LGTM with some suggestions.

update/src/workflows.ts
dandavison: I thought about mentioning that in the comment but decided not to, because I had heard you say that in your opinion async update handlers should be considered an advanced use case only and need to be used with care. So it seemed best to keep this introductory sample simple.

bergundy: Maybe it's worth doing another sample with an async update function when we have update-with-start for the async compensation pattern.

PR #337, opened 2024-02-09
Adding missing @temporalio/common dependency for expenses example

emadgit:
## What was changed

it seems `@temporalio/common` dependency is missing for expenses example.
## Why?

Because it is imported and used in `expense/src/workflows.ts` . This PR adding this dependency to package.json


mjameswh (approved):
Thanks!

PR #342, opened 2024-02-23
Update mutex sample to use workflowInfo().continueAsNewSuggested

HalfdanJ:
## What was changed
Instead of relying on history length to determine when to restart workflow, utilize continueAsNewSuggested on workflowInfo. 

## Why?
This feature was showcased yesterday as a more accurate approach at Temporal meetup by @lorensr  

## Checklist
1. Closes: n/a

2. How was this tested:
Running the unit tests 

3. Any docs updates needed?
No


lorensr:
Great, thanks!

PR #343, opened 2024-02-23
Single file example

dandavison:
From README:

> This sample demonstrates a minimalistic single-file style that might be useful for quickly sketching out an idea, or for sharing a minimal reproduction of a bug.
Production code should not be arranged like this! See the other samples in this repo for recommended project structure.


hello-world-single-file/README.md
```diff
@@ -0,0 +1,15 @@
+# Hello World (single-file)
```
Sushisource, line 1: I would call this whole sample scratchpad - I think that can be a really useful way to present it to users. "Hey, you want a simple way to develop everything all together while you're experimenting? Here you go."

dandavison: Nice, OK I've renamed it as "scratchpad".

Sushisource (approved):
Might be worth adding a sentence or two about why this is different from many of the other samples, and how typically in production you will have your client code living elsewhere, etc.

scratchpad/scratchpad.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function exampleWorkflow(name: string): Promise<string> {
+  return await greetActivity(name);
+}
+
+async function starter(client: Client): Promise<void> {
+  const result = await client.workflow.execute(exampleWorkflow, {
+    taskQueue,
+    workflowId: 'my-workflow-id',
+    args: ['Temporal'],
+  });
```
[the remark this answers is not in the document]
dandavison: Yes, good call. I think I might also the ability to pass the workflow ID in as a command-line argument before merging. That will be helpful to someone using this as a template for something where they want command line arguments.

scratchpad/README.md
```diff
@@ -0,0 +1,24 @@
...
+In one terminal:
+
+```
+temporal server start-dev
+```
+
+In another terminal:
+
+```
+ts-node scratchpad.ts
```
lorensr, line 17:
```suggestion
ts-node scratchpad.ts -w my-workflow-id -t my-task-queue
```

lorensr (approved):
Nice! 😃🙏

Add a custom

[.post-create](https://github.com/temporalio/samples-typescript/blob/main/.shared/.post-create)

and include in:

https://github.com/temporalio/samples-typescript/blob/5a984226ff687f49b6fa1aa74ea209f02d9b3ecc/.scripts/copy-shared-files.mjs#L51

scratchpad/README.md
dandavison: Thanks. The CLI arg parsing was an experiment. I've decided to revert it, seeing as the point here is minimalism. So the workflow ID and task queue are top-level constants (and the workflow uses `WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING`).

PR #344, opened 2024-02-23
Correct the mutex sample docs

tdeebswihart:
## What was changed

I updated the README for our Mutex workflow sample

## Why?

The usage details and example output were incorrect


lorensr:
Thanks! ☺️

PR #347, opened 2024-02-26
Get scratchpad working with create script

lorensr:
Currently, `npx @temporalio/create@latest my-project --sample scratchpad` isn't working. Added a `package.json` and got the commit hook to run on it and `update` to add the missing files.


scratchpad/.post-create
```diff
@@ -7,9 +7,11 @@ Start Temporal Server:
 
   {cyan temporal server start-dev}
 
-Install ts-node: https://www.npmjs.com/package/ts-node
+Use Node version 16+:
```
dandavison, line 10: Can we change this to a more recent node version?

scratchpad/README.md
```diff
@@ -5,18 +5,9 @@ Production code should not be arranged like this! See the other samples in this
...
-
-In another terminal:
-
-```
-npm install temporalio
-ts-node scratchpad.ts
-```
+1. `temporal server start-dev` to start [Temporal Server](https://github.com/temporalio/cli/#installation).
+1. `npm install` to install dependencies.
+1. `npm run start.watch` to run the single file, [`scratchpad.ts`](scratchpad.ts).
```
dandavison, line 10: My idea for the "scratchpad" collection of examples is to have all of them runnable using the simplest possible command-line techniques, i.e. so that they run, print the workflow result to stdout, and exit (without info-level logging). So what I was thinking so far is that this would be

```
python scratchpad.py
go run scratchpad.go
./gradlew -q execute -PmainClass=io.temporal.samples.hello.Scratchpad  # really java?
ts-node scratchpad.ts
# dotnet TODO
```

In addition, my idea was to not require special scaffolding to be set up (e.g. package.json).

So I really like the fact that with `scratchpad.ts` someone can just do

```
curl -sOL https://raw.githubusercontent.com/temporalio/samples-typescript/main/scratchpad/scratchpad.ts
npm i temporalio
ts-node scratchpad.ts
```

scratchpad/.post-create
lorensr: i think they'll all get changed together soon

lorensr: (looks like involves changing the create tool in sdk-typescript too)

scratchpad/README.md
lorensr: these samples are meant to be used by the create tool, which atm requires a package.json. that could be modified, or i could just add a section above this, maybe:

```
### Quickstart

curl -sOL https://raw.githubusercontent.com/temporalio/samples-typescript/main/scratchpad/scratchpad.ts 
npm i -g temporalio 
npm i -g ts-node
ts-node scratchpad.ts

### Watch changes

<current Running this sample section>
```

?

dandavison: I like that!

dandavison: (You know `npm` better than me but I notice that `-g` is not necessary even when there is no package.json; is `-g` implied in that situation? If so perhaps it's better to omit the `-g` from the instructions so that they install into local `node_modules` for someone who has in fact cloned the repo rather than `curl -sOL`?)

lorensr: for me, without `-g` it does this:

image

will clarify it's for if you didn't clone

dandavison: Yes that's what it does for me. That's OK isn't it? Creates a local package rather than modifying the user's global environment.

lorensr: oh, i was thinking it's a nuisance to have these extra files in whatever location the user is in. thinking they may not have created a directory for this. and if the dev is concerned about global modules they'll know what the -g means and remove it? i'm pretty used to -g in my dev experience, but i'll check with folks in slack

lorensr: everyone else says not global 😊 updating!

PR #360, opened 2024-04-18
fix workflowtype stripped

antmendoza:
## What was changed

 // When the project is built, the bundler will strip out Workflow function names.
  // This ensures the workflow type remains the same
  // when the client send the request to Temporal Server. Another option is modifying the webpack config,
  // as it is stated here https://docs.temporal.io/dev-guide/typescript/debugging#production-bundling

Another option, I guess, is to pass an string as workflowType 

```
await client.workflow.start(`OneClickBuy`, {
```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


antmendoza:
@mjameswh   I show your comment in Jira, let me modify this accordingly

PR #363, opened 2024-05-12
Fixed fakeProgress activity heartbeat

ssrohit:
## What was changed
Changed activity(long-running activity) to heartbeat immediately 

## Why?
If the activity heartbeats after operation such as DB insertion or reads, when failed the activity retries from a progress point which is success, which results in inconsistency. 

1. Closes - https://github.com/temporalio/samples-typescript/issues/366

2. How was this tested:
I have tested this in my personal project, one can test and replicate this by data insertion in the fakeProgress activity instead of `await sleep(...)` code line.

3. Any docs updates needed?
Yes, code sample in the documentation should be updated and this should be described for better understanding.
https://docs.temporal.io/dev-guide/typescript/features#activity-heartbeats


ssrohit:
This PR can close this issue https://github.com/temporalio/samples-typescript/issues/366

PR #365, opened 2024-05-15
Next.js Oneclick: Reuse Temporal Connection across API invocations

### LINKED ISSUE #322: Typo in `nextjs-ecommerce-oneclick` example?
It seems like this line should use `Client()`, not `Workflow()`?
```ts
 const client = new Workflow();
```
https://github.com/temporalio/samples-typescript/blob/f16fa9d4962c31231ea63c0a7c7dada3c52a840a/nextjs-ecommerce-oneclick/pages/api/getBuyState.js#L10C33-L10C33

mjameswh:
## What was changed and why?

- Move Temporal Connection and Client initialization to a distinct module
  - Avoids creating a new Connection on each API calls. gRPC connections are heavyweight and show be reused as much as possible.

- Fixed the UI side request state machine
  - It was masking failed API calls, resulting in incorrect "Purchased!" messages.

## Misc
- Closes #322

### DIFF
--- a/nextjs-ecommerce-oneclick/pages/api/cancelBuy.js
+++ b/nextjs-ecommerce-oneclick/pages/api/cancelBuy.js
@@ -1,4 +1,4 @@
-import { Client } from '@temporalio/client';
+import { getTemporalClient } from '../../temporal/lib/client.js';
 
 export default async function cancelBuy(req, res) {
   const { id } = req.query;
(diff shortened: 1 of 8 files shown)


bergundy (approved):
Not sure you should put `getTemporalClient` in `temporal/lib` if it's only used in these routes but 🤷

mjameswh:
> Not sure you should put `getTemporalClient` in `temporal/lib` if it's only used in these routes but 🤷

Anything in the `/api/` directory becomes a route. Anything internal/utility goes outside the `/pages` subtree.

PR #369, opened 2024-05-31
Fix comments in grpc sample

bergundy:


grpc-calls/src/client.ts
```diff
@@ -18,7 +18,7 @@ async function run() {
     // This should not happen with standard inputs and the defaultPayloadConverter.
     throw new TypeError('Could not convert string to payload');
   }
-  // equivalent grpc call to client.start()
+  // equivalent grpc call to client.workflow.start()
```
mjameswh, line 21: None of the snipsync snippets in this file is still referenced. Would you mind removing them all before merging?

mjameswh (approved):

grpc-calls/src/client.ts
bergundy: done

PR #372, opened 2024-07-03
Message handlers sample

dandavison:
Fixes https://github.com/temporalio/sdk-typescript/issues/1446

This is based on @drewhoskins-temporal's Python sample: https://github.com/temporalio/samples-python/pull/123, and uses https://github.com/DirtyHairy/async-mutex.

Note that this moves husky actions to pre-push; they are too disruptive in pre-commit for people who commit frequently, including via non-CLI clients that would have to be configured to pass `--no-verify`.


updates_and_signals/safe_message_handlers/src/run-simulation.ts
```diff
@@ -0,0 +1,36 @@
...
+  await wf.signal(startClusterSignal);
+
+  const allocationUpdates: Promise<any>[] = [];
+  for (let i = 0; i < 6; i++) {
+    allocationUpdates.push(wf.executeUpdate(assignNodesToJobUpdate, { args: [{ numNodes: 2, jobName: `task-${i}` }] }));
+  }
+  await Promise.all(allocationUpdates);
+
+  if (delaySeconds) {
+    await new Promise((resolve) => setTimeout(resolve, delaySeconds * 1000));
```
mjameswh, line 16:
```
import { setTimeout } from 'timers/promises';
await setTimeout(delaySeconds * 1000);
```

updates_and_signals/safe_message_handlers/src/workflows.ts
```diff
@@ -0,0 +1,63 @@
...
+  //
+  // The cluster manager workflow is a long-running workflow ("entity" workflow). Most of its logic
+  // lies in the message-processing handlers implented in the ClusterManager class. The main
+  // workflow itself is a loop that does the following:
+  // - process messages
+  // - continue-as-new when suggested
+  //
+  await wf.condition(() => manager.state.clusterStarted);
+  for (;;) {
+    await wf.condition(() => manager.state.clusterShutdown || wf.workflowInfo().continueAsNewSuggested);
```
mjameswh, line 54: DANGER!!! It's possible for `continueAsNewSuggested` to turns true while there are pending async operations in cluster manager.

dandavison: Ah of course, thanks!

updates_and_signals/safe_message_handlers/src/activities.ts
```diff
@@ -0,0 +1,28 @@
...
+  console.log(`Assigning nodes ${input.nodes} to job ${input.jobName}`);
+  await sleep(100); // Simulate RPC
+}
+
+export async function unassignNodesForJob(input: UnassignNodesForJobInput): Promise<void> {
+  console.log(`Unassigning nodes ${input.nodes} from job ${input.jobName}`);
+  await sleep(100); // Simulate RPC
+}
+
+async function sleep(ms: number): Promise<void> {
```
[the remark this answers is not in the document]
dandavison: Thanks, done!

updates_and_signals/safe_message_handlers/src/run-simulation.ts
dandavison: done!

updates_and_signals/safe_message_handlers/src/workflows.ts
dandavison: Changed to wait on `allhandlersFinished`

dandavison:
@mjameswh I think I've addressed everything here. Over to you again.

updates_and_signals/safe_message_handlers/src/client.ts
```diff
@@ -0,0 +1,14 @@
+import { Connection, Client, WorkflowHandle } from '@temporalio/client';
+import { v4 as uuid } from 'uuid';
+
+import { clusterManagerWorkflow } from './workflows';
+
+export async function startClusterManager(): Promise<WorkflowHandle<typeof clusterManagerWorkflow>> {
+  const connection = await Connection.connect({ address: 'localhost:7233' });
+  const client = new Client({ connection });
+  return client.workflow.start(clusterManagerWorkflow, {
+    args: [{}],
```
mjameswh, line 10: nit: Use a default arg value on the `clusterManagerWorkflow` function instead, and no arg here. The fact that the workflow propagates its state through this arg on CAN is an internal detail that client-side has no need to know in this case.

updates_and_signals/safe_message_handlers/src/client.ts
```diff
@@ -0,0 +1,14 @@
...
+    taskQueue: 'safe-message-handlers-task-queue',
+    workflowId: `cm-${uuid().slice(0, 8)}`,
```
mjameswh, line 12: Why this `slice`? Users will wonder if that's a best practice or something they should replicate in their own code.

updates_and_signals/safe_message_handlers/src/cluster-manager.ts
```diff
@@ -0,0 +1,134 @@
...
+import { Mutex } from 'async-mutex';
+import {
+  AssignNodesToJobUpdateInput,
+  ClusterManagerState,
+  ClusterManagerStateSummary,
+  DeleteJobUpdateInput,
+} from './types';
+
+const { assignNodesToJob, unassignNodesForJob, startCluster } = wf.proxyActivities<typeof activities>({
+  startToCloseTimeout: '1 minute', // TODO
```
mjameswh, line 12: Left over TODO

updates_and_signals/safe_message_handlers/src/cluster-manager.ts
```diff
@@ -0,0 +1,134 @@
...
+    await wf.condition(() => this.state.clusterStarted);
+    if (this.state.clusterShutdown) {
+      // If you want the client to receive a failure, either add an update validator and throw the
+      // exception from there, or raise an ApplicationError. Other exceptions in the handler will
+      // cause the workflow to keep retrying and get it stuck.
+      throw new wf.ApplicationFailure('Cannot assign nodes to a job: Cluster is already shut down');
+    }
+    return await this.nodesMutex.runExclusive(async (): Promise<ClusterManagerStateSummary> => {
+      // Idempotency guard: do nothing if the job already has nodes assigned.
+      if (!new Set(this.state.nodes.values()).has(input.jobName)) {
```
mjameswh, line 64: What if the job has already been assigned and completed? Do we care?

updates_and_signals/safe_message_handlers/src/cluster-manager.ts
```diff
@@ -0,0 +1,134 @@
...
+      // exception from there, or raise an ApplicationError. Other exceptions in the handler will
+      // cause the workflow to keep retrying and get it stuck.
+      throw new wf.ApplicationFailure('Cannot delete job: Cluster is already shut down');
+    }
+    await this.nodesMutex.runExclusive(async () => {
+      const nodesToUnassign = Array.from(this.state.nodes.entries())
+        .filter(([_, v]) => v === input.jobName)
+        .map(([k, _]) => k);
+      // This await would be dangerous without the lock held because it would allow interleaving
+      // with the assignNodesToJob and performHealthCheck operations, both of which mutate
```
mjameswh, line 97: `performHealthCheck`? Looks like that doesn't exist anymore.

updates_and_signals/safe_message_handlers/src/cluster-manager.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  getUnassignedNodes(): Set<string> {
+    return new Set(Array.from(this.state.nodes.keys()).filter((key) => this.state.nodes.get(key) === null));
+  }
+
+  getAssignedNodes(jobName?: string): Set<string> {
+    return new Set(
+      Array.from(this.state.nodes.keys()).filter((key) => {
+        const value = this.state.nodes.get(key);
+        return jobName ? value === jobName : value !== null && value !== 'BAD!';
```
mjameswh, line 130: What's the expected behavior if `jobName` is an empty string? `Boolean("")` is `false`.

updates_and_signals/safe_message_handlers/src/test/workflows.test.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as activities from '../activities';
+import {
+  clusterManagerWorkflow,
+  assignNodesToJobUpdate,
+  startClusterSignal,
+  shutdownClusterSignal,
+  deleteJobUpdate,
+  getClusterStatusQuery,
+} from '../workflows';
+import { nanoid } from 'nanoid';
```
mjameswh, line 13: Why use `nanoid` here but `uuid` before? I'd suggest keeping `nanoid`.

updates_and_signals/safe_message_handlers/src/test/workflows.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  after(async function () {
+    await env.teardown();
+  });
+
+  it('successfully completes a session', async function () {
+    await worker.runUntil(async function () {
+      const workflow = await env.client.workflow.start(clusterManagerWorkflow, {
+        args: [{}],
```
mjameswh, line 50: As said before, set default arg value at the function level instead of explicitly passing this empty object.

updates_and_signals/safe_message_handlers/src/test/workflows.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        taskQueue,
+        workflowId: 'cluster-manager-' + nanoid(),
```
mjameswh, line 52:
```suggestion
        workflowId: `cluster-manager-${ nanoid() }`,
```

updates_and_signals/safe_message_handlers/src/workflows.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  //
+  // Main workflow logic
+  //
+  // The cluster manager workflow is a long-running workflow ("entity" workflow). Most of its logic
+  // lies in the message-processing handlers implented in the ClusterManager class. The main
+  // workflow itself simply waits until the cluster is shutdown, or the workflow needs to
+  // continue-as-new.
+  await wf.condition(() => manager.state.clusterShutdown || wf.workflowInfo().continueAsNewSuggested);
+  if (wf.workflowInfo().continueAsNewSuggested) {
```
mjameswh, line 50: If `clusterShutdown === true`, then we should imperatively take the shutdown path, whatever the value of `continueAsNewSuggested`.

mjameswh (changes requested):
Left multiple comments

updates_and_signals/safe_message_handlers/src/cluster-manager.ts
dandavison: Thanks. Switched to store seen job names in a set. This is what our Python version does. However it looks like our .NET sample has no idempotency guard cc @cretz.

dandavison (commented):
Thanks for the careful reviewing @mjameswh! Addressed those.

PR #382, opened 2024-09-09
Sample for message passing docs

dandavison:
This PR adds sample code used in the message-passing docs: https://github.com/temporalio/documentation/pull/3070


updates-and-signals/introduction/src/workflows.ts
```diff
@@ -0,0 +1,126 @@
...
+export const getLanguages = wf.defineQuery<Language[], [GetLanguagesInput]>('getLanguages');
+export const approve = wf.defineSignal<[ApproveInput]>('approve');
+export const setLanguage = wf.defineUpdate<Language, [Language]>('setLanguage');
+export const setLanguageUsingActivity = wf.defineUpdate<Language, [Language]>('setLanguageUsingActivity');
+export const getLanguage = wf.defineQuery<Language>('getLanguage');
+
+export async function greetingWorkflow(): Promise<string> {
+  // The workflow has built-in support for two languages only (see remote
+  // greeting service below for support for more languages).
+  const greetings: Partial<Record<Language, string>> = {
```
dandavison, line 33: @mjameswh do you see a better way to type this? The idea is that this object holds translations that are available in-workflow, and that the activity can be used to obtain other translations. I hesitated to create the object with most values equal to empty-string, since the code is used as a docs snippet and it looked a bit distracting.

message-passing/introduction/src/workflows.ts
```diff
@@ -0,0 +1,127 @@
...
+    // 👉 A Signal handler mutates the Workflow state but cannot return a value.
+    approvedForRelease = true;
+    approverName = input.name;
+  });
+
+  wf.setHandler(setLanguage, (newLanguage) => {
+    // 👉 An Update handler can mutate the Workflow state and return a value.
+    if (!(newLanguage in greetings)) {
+      throw new Error(`${newLanguage} is not supported`);
+    }
```
dandavison, line 62: This should be a validator; fixing

dandavison: Fixed

message-passing/introduction/src/test/workflows.test.ts
```diff
@@ -0,0 +1,153 @@
+import { TestWorkflowEnvironment } from '@temporalio/testing';
+import { before, describe, it } from 'mocha';
+import * as wo from '@temporalio/worker';
+import * as greetingWorkflow from '../workflows';
+import { Language } from '../workflows';
+import assert from 'assert';
+const taskQueue = 'test' + new Date().toLocaleDateString('en-US');
```
mjameswh, line 7: That's useless, since you're using a `TestWorkflowEnvironment` anyway.

mjameswh (approved):
Looks good. Thanks!

message-passing/introduction/src/test/workflows.test.ts
dandavison: Thanks, fixed (that was copied from other test files in the repo).

PR #386, opened 2024-10-04
Fix cluster start/stop

dandavison:
- Use an update for `clusterStop`: fixes bug wherein `stop` sent while cluster is down would hang until it comes up, then take it down
- Make `clusterStart` idempotent: fixes bug wherein a `start` sent while the cluster was up would overwrite the nodes data
- Do `clusterStart` with lock held (otherwise an assign job could come in while it's initializing
- Use an enum for cluster state instead of boolean flags

### DIFF
--- a/message-passing/safe-message-handlers/src/cluster-manager.ts
+++ b/message-passing/safe-message-handlers/src/cluster-manager.ts
@@ -40,11 +40,13 @@ export class ClusterManager {
     if (this.state.clusterState === ClusterState.UP) {
       return;
     }
-    await startCluster();
-    this.state.clusterState = ClusterState.UP;
-    for (let i = 0; i < 25; i++) {
-      this.state.nodes.set(i.toString(), null);
-    }
+    await this.nodesMutex.runExclusive(async () => {
+      await startCluster();
+      this.state.clusterState = ClusterState.UP;
+      for (let i = 0; i < 25; i++) {
+        this.state.nodes.set(i.toString(), null);
+      }
+    });
     wf.log.info('Cluster started');
   }
 
--- a/message-passing/safe-message-handlers/src/cluster-manager.ts
+++ b/message-passing/safe-message-handlers/src/cluster-manager.ts
@@ -4,11 +4,12 @@ import { Mutex } from 'async-mutex';
 import {
   AssignNodesToJobUpdateInput,
   ClusterManagerState,
+  ClusterState,
   ClusterManagerStateSummary,
   DeleteJobUpdateInput,
 } from './types';
 
-const { assignNodesToJob, unassignNodesForJob, startCluster } = wf.proxyActivities<typeof activities>({
+const { assignNodesToJob, unassignNodesForJob, startCluster, shutdownCluster } = wf.proxyActivities<typeof activities>({
   startToCloseTimeout: '1 minute',
 });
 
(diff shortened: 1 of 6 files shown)


message-passing/safe-message-handlers/src/types.ts
```diff
@@ -28,3 +27,8 @@ export interface ClusterManagerWorkflowResult {
   numCurrentlyAssignedNodes: number;
   numBadNodes: number;
 }
+
+export enum ClusterState {
+  UP = 'UP',
```
Quinn-With-Two-Ns, line 32: I think we needed a 3rd state here for "not started" otherwise start isn't idempotent.I think the state machine should be like

NOT STARTED -> STARTED -> CLOSED right now it is  STARTED <-> CLOSED

dandavison: > I think we needed a 3rd state here for "not started" otherwise start isn't idempotent

OK, what's the bad behavior you're seeing exactly? `startCluster` is idempotent currently, right -- after applying the operation once, from either initial state, you enter an infinite sequence of `UP` states:

image

Quinn-With-Two-Ns: In your code you can transition for `Up` to `Down` in `shutDownCluster`. That

dandavison: Right, so the state machine is like this; both operations are idempotent. I'm still not getting the problem:

image

Quinn-With-Two-Ns: you can run `startCluster` ,... other messages , and then `shutDownCluster` and `startCluster`. I think you will clear the state of the nodes

dandavison: Right, so the cluster completely forgets its state when it's shut down. It's always initialized when it starts up

```typescript
      this.state.clusterState = ClusterState.UP;
      for (let i = 0; i < 25; i++) {
        this.state.nodes.set(i.toString(), null);
      }
```

(the number of nodes 25 never changes). That's OK isn't it, for a toy example?

dandavison: Discussed offline. The issue was that I was using the math/CS definition of idempotent whereas in the context of Temporal activities the definition is more like "will never happen a second time having happened once". Agree it makes sense to make the operations idempotent in the Temporal Activityh sense; I've switched to a 3-state state space now. Thanks, interesting discussion!

Quinn-With-Two-Ns (approved):

message-passing/safe-message-handlers/src/cluster-manager.ts
```diff
@@ -37,27 +37,39 @@ export class ClusterManager {
...
+    await shutdownCluster();
+    this.state.clusterState = ClusterState.SHUTTING_DOWN;
     wf.log.info('Cluster shutdown');
+    return true;
   }
 
   async assignNodesToJob(input: AssignNodesToJobUpdateInput): Promise<ClusterManagerStateSummary> {
-    await wf.condition(() => this.state.clusterStarted);
-    if (this.state.clusterShutdown) {
+    await wf.condition(() => this.state.clusterState === ClusterState.STARTED);
```
Quinn-With-Two-Ns, line 67: @dandavison I just noticed is there a deadlock here if the `clusterState` goes from `STARTED` to `SHUTTING_DOWN` without this handler running?

PR #393, opened 2024-11-21
Update NestJS sample

MBach:
## What was changed
- The NestJS sample was rewritten in a more "real world" style
- Readme was updated
- Webpack for production, default (tsc) for development

## Why?
- NPM scripts weren't working properly
- Repository was using outdated NestJS framework


mjameswh (approved):
Thanks a lot, @MBach. C'est très apprécié!

PR #395, opened 2024-11-29
Close all explicitly created NativeConnection

mjameswh:
## What was changed

- Add a finally block containing `await connection.close()` for all native connections that are explicitly created from the code. Not doing so prevents the worker process from exiting.


dandavison (approved):
LGTM!

PR #396, opened 2024-12-03
Upgrade to TS SDK 1.11.5 / Node 18 to 22

mjameswh:
## What was changed

- Upgraded to TS SDK 1.11.5
- Upgraded supported Node versions to 18..22 (instead of 16..20)


yuandrew (approved):
Didn't take a super in-depth look, but from a glance everything looks good

PR #398, opened 2024-12-03
Convert workspace to PNPM

mjameswh:
## What was changed

- Convert the root project to PNPM, instead of Lerna + Yarn Classic.

  - The main motive for this change is to provide proper isolation of dependencies between samples; `npm`'s dependency hoisting was previously causing various issues due to incompatible dependencies version requirements, making it very hard to maintain the repo up-to-date.

  - As bonus, this change also significantly reduce the time it takes to install all dependencies in the repo, as well as the disk space required.

- Modified CI workflow to use a single job to check lint on all projects.

- Update TS dependencies to 4.9.5 (just a temporary measure to make it easier to synchronize tsconfig everywhere, this will be bumped to 5.6 in a distinct PR).

- Renamed packages of the Food Delivery sample to `@fooddelivery/*` to clarify dependencies.

- Remove `turbo` in the Food Delivery sample, because we don't really care about the cache optimization it provides in this specific case.


dandavison (approved):
Excellent! Happy to see improvements here. Do any of the README instructions need to change?

PR #401, opened 2024-12-11
Update-with-start early return sample

dandavison:
Cannot be merged until UwS is released.


early-return/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "ext": "ts",
+    "watch": [
+      "src"
+    ]
+  },
+  "dependencies": {
+    "@temporalio/activity": "file:~/src/temporalio/sdk-typescript/packages/activity",
+    "@temporalio/client": "file:~/src/temporalio/sdk-typescript/packages/client",
+    "@temporalio/worker": "file:~/src/temporalio/sdk-typescript/packages/worker",
+    "@temporalio/workflow": "file:~/src/temporalio/sdk-typescript/packages/workflow",
```
dandavison, line 27: TODO: remove local changes

early-return/src/workflows.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getTransactionConfirmation, TransactionReport } from './shared';
+
+export async function transactionWorkflow(): Promise<TransactionReport> {
+  let confirmed = false;
+  wf.setHandler(getTransactionConfirmation, async () => {
+    await wf.condition(() => confirmed);
+    const status = 'confirmed' as const;
+    return { status };
+  });
+  await wf.sleep(1000);
```
mjameswh, line 11: As it is now, both the Update and the Workflow completes in the same WFT. You should have a sleep _after_ `confirmed = true` (e.g. before completing the Workflow). I have no opinion though on either it is pertinent to have a `sleep` _before_ confirming the update.

early-return/src/start-workflow.ts
```diff
@@ -0,0 +1,34 @@
...
+    workflowId,
+    taskQueue: 'early-return',
+    workflowIdConflictPolicy: 'FAIL',
+  });
+
+  const earlyConfirmation = await client.workflow.executeUpdateWithStart(getTransactionConfirmation, {
+    startWorkflowOperation,
+  });
+  const wfHandle = await startWorkflowOperation.workflowHandle();
+  const finalReport = await wfHandle.result();
```
mjameswh, line 18: You should print the early confirmation before awaiting for the workflow completion. Right now, both will display at the same time, which fails to demonstrate the notion of an "early return".

early-return/src/workflows.ts
mjameswh: Also, I would make the sleep duration between update completion and wf completion to be something significant… Let's say 5, or even 10 seconds. And add a comment on that sleep saying would be be some long operation that can be done asynchrnously…

early-return/src/start-workflow.ts
dandavison: Done

early-return/src/workflows.ts
dandavison: Good call, done.

mjameswh (approved):
Only did a quick last review, but that looks ok to me. If anything pops up, we'll revisit.

Thanks @dandavison !

PR #402, opened 2024-12-18
fix(tests): Avoid Node v22.12.0 on Windows

mjameswh:
## What was changed

- In CI, avoids Node v22.12.0 on Windows, because [it incorrectly resolves `localhost` to `::1`](https://github.com/nodejs/node/issues/56137), rather than both `::1` and `127.0.0.1`, which is causing errors when executing tests. We can remove this workaround once a fix version of Node is released.


dandavison (approved):
LGTM

PR #405, opened 2025-01-02
fix: added husky to package.json in nextjs ecommerce app

GSmithApps:
## What was changed
Husky was added to the NextJS Ecommerce App package.json dev dependencies

## Why?
Without this here, I got an error that said I need husky.

## How was this tested
I ran `npm install` after adding this, and the error went away.

### DIFF
--- a/nextjs-ecommerce-oneclick/package.json
+++ b/nextjs-ecommerce-oneclick/package.json
@@ -41,6 +41,7 @@
     "eslint-config-next": "^15.0.4",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-deprecation": "^3.0.0",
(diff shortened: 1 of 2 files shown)


mjameswh:
Thanks, Grant!

We recently converted the root project to PNPM, which already declares that dependency, but all samples are expected to be runnable in isolation, with either `npm`, `pnpm` or `yarn` unless explicitly indicated, so this change is very pertinent. Thanks again.

mjameswh (approved):

GSmithApps:
Perfect - thank you very much, James! Yeah switching that package manager probably causes quite a kerfuffle. I'm impressed the team has it working so well already 👍🏼 Cheers

PR #408, opened 2025-01-16
Revise coverage of API Key authentication

tomwheeler:
## What was changed

I changed the name of the environment variable used to reference the API key from `TEMPORAL_CLIENT_KEY_PATH` to `TEMPORAL_CLIENT_API_KEY `. I also updated the `README.md` to better reflect the endpoint addresses currently shown for mTLS and API keys in Temporal Cloud. Finally, I updated the two source files that establish a connection to the Temporal Service so that they set the `apiKey` attribute and metadata, as per [the documentation](https://docs.temporal.io/cloud/api-keys#typescript-sdk-v-1100) for post-1.10.0 releases of the TypeScript SDK.

## Why?

This makes the code follow what's documented in the README, which specifies the `TEMPORAL_CLIENT_API_KEY` environment variable for the API key. The `TEMPORAL_CLIENT_KEY_PATH` environment variable is already used to specify the location of the TLS private key, which leads me to believe that its reuse on line 32 was a mistake.

Also, the explanation in the README about what endpoint addresses look like did not reflect those shown in Temporal Cloud. My change updates the description to match what I see in the UI.

Finally, my connections to Temporal Cloud were rejected as unauthorized because the code wasn't configured correctly for a Client using API Key for authentication.

## Checklist

1. Closes: N/A

2. How was this tested: I tested both the mTLS and API key authentication scenarios using my Temporal Cloud account.

### DIFF
--- a/hello-world-mtls/README.md
+++ b/hello-world-mtls/README.md
@@ -60,7 +60,7 @@ npm run workflow to run the Workflow
 - **Address**: Make sure to configure the mTLS endpoint as it appears in Temporal Cloud's web interface,
   e.g. `${namespace}.tmprl.cloud:7233`.
 
-  Note that endpoints of the form `${region}.${provider}.api.tmprl.cloud:7233` only support [API Key authentication](#connection-to-a-temporal-cloud-namespace-with-api-key-authentication). They will not work with mTLS authentication.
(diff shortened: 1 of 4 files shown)


mjameswh (approved):
That all totally makes sense. Thanks a lot!

PR #410, opened 2025-01-24
Do not configure NPM workspace on the root project

mjameswh:
## What was changed

- Do not configure NPM workspace on the root repository. We specifically want
  PNPM and NPM/YARN configuration to differ on this regard, so that the repository
  can be used either as a monorepo (e.g. `pnpm install` on the root directory) or
  as individual projects (e.g. `cd hello-world ; npm install`, without installing
  dependencies on the whole repo).

  That's how the PNPM setup was initially created. I added that at some point to
  test something, and may have I forgot to remove before the final commit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,12 +35,5 @@
         }
       }
     }
-  },
-  "workspaces": [
-    "*",
-    "message-passing/*",
-    "monorepo-folders/packages/*",
-    "food-delivery/packages/*",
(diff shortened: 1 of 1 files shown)


GSmithApps:
Commenting here so GitHub sees me as "involved" in this and I can find it later and I get notifications etc.

Apologies for the extra notification, James 🙏

PR #412, opened 2025-01-25
Add `sleep-for-days` sample

THardy98:
This is a simple sample demonstrating Temporal's ability to sleep for days and still execute reliably.

## What was changed
added `sleep-for-days` sample

How was this tested:
added a test inside the sample, see `test/workflows.test.ts`

3. Any docs updates needed?
I believe so @webchick


sleep-for-days/src/workflows.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/** A workflow that repeatedly calls an activity and sleeps, until it receives a signal. */
+export async function sleepForDays(numDays: number): Promise<string> {
+  let isComplete = false;
+  setHandler(complete, () => { isComplete = true });
+  
+  while (!isComplete) {
+    await sendEmail(`Sleeping for ${numDays} days`);
+    await sleep(days(numDays));
+  }
```
Sushisource, line 22: Same as my comment in the python PR - the loop should race against the completion

THardy98: Done.

sleep-for-days/src/workflows.ts
```diff
@@ -0,0 +1,28 @@
...
+/** A workflow that repeatedly calls an activity and sleeps, until it receives a signal. */
+export async function sleepForDays(numDays: number): Promise<string> {
+  let isComplete = false;
+  setHandler(complete, () => { isComplete = true });
+  
+  while (!isComplete) {
+    await sendEmail(`Sleeping for ${numDays} days`);
+    await Promise.race([
+      sleep(days(numDays)),
+      condition(() => isComplete),
```
dandavison, line 23: Same as Python -- we should be able to use the `timeout` parameter of `condition(...)` instead of explicitly racing a promise.

THardy98: Yeah, that's true. I talked about it with Spencer and I think the intention of the sample is show something easily readable, and `sleep` is a bit more intuitive.

I wrote out this comment on the Python PR but left it in draft

Sushisource: Yeah - this being mainly to get people curious on the website, I want them to see a "sleep" that says 30 days and be like "oh wow cool how does that work?", and as a parameter to condition I'm worried it doesn't have that same effect.

sleep-for-days/src/workflows.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/** A workflow that repeatedly calls an activity and sleeps, until it receives a signal. */
+export async function sleepForDays(numDays: number): Promise<string> {
+  let isComplete = false;
+  setHandler(complete, () => { isComplete = true });
+  
+  while (!isComplete) {
+    await sendEmail(`Sleeping for ${numDays} days`);
+    await Promise.race([
+      sleep(days(numDays)),
```
Sushisource, line 22: We don't need the `days` helper since this accepts [formatted strings](https://www.npmjs.com/package/ms). You can just write `${numDays} days`

sleep-for-days/src/workflows.ts
dandavison: OK cool, I buy this.

May I make another suggestion: how about we get rid of the parameterization and just have this literally say `sleep(days(30))`? The point of the sample is that they can do that, not that they can choose the precise length of the really long time dynamically from the caller.

Sushisource: I'm a fan of that too.

THardy98: Change it to have no params.

I kept it with a param in the beginning because I thought it'd be easier to test, but seeing as we have time-skipping (at least in the Typescript SDK) should be a non-issue.

Interested to hear your thoughts on how to test in other SDKs that don't (easily) support manual time-skipping in test environments. In this case, would it be easiest to keep the param and sub it for some very low value in tests (i.e. in the order of seconds)?

sleep-for-days/src/workflows.ts
THardy98: Done.

sleep-for-days/src/workflows.ts
dandavison: In Python I wonder if you could `mock.patch` `asyncio.sleep` so that it does call the original, but with a short duration? It's possible this would require `UnsandboxedWorkflowRunner` -- I'd have to experiment.

dandavison: That said, which SDKs does this describe?

> SDKs that don't (easily) support manual time-skipping in test environments

THardy98: Honestly, I'm not sure. I started this conversation with James but ran out of time before sprint planning, so never really got into it. But what I gathered was each SDK can have testing support for both a CLI dev server and time-skipping dev server. I'm assuming a CLI dev server is a bit of an older pattern (where the tests by necessity need a CLI server to run, I've seen this pattern in the TS SDK with older tests) and the time-skipping dev server is a newer development that's baked into the code test environment.

From that perspective, this is what each SDK supports (to my understanding):

Go SDK: only supports the CLI dev server
Java SDK: only supports the time skipping server
Core-based SDKs: support both servers through core

I very well could be missing something (and probably am), but that's my lay of the land. Anecdotally, this seems to follow. I couldn't find a way to do time-skipping with the Go SDK (at least from the code).

One thing that I'm not sure about is in the Python SDK I see:
```
if env.supports_time_skipping:
        pytest.skip(
            "Java test server: https://github.com/temporalio/sdk-java/issues/1903"
        )
```
which makes me think time-skipping is a no-no but I think I just don't really understand what's going on there (or rather, how time-skipping in tests are possible in the Python SDK if we use a conditional that explicitly avoids time-skipping).

Sushisource: Well, that snipped you posted links to a specific bug in the time skipping server. It doesn't mean time skipping is a problem generally.

Lastly - none of this matters really. If time skipping works, great, use `30 days`. If it doesn't, the sample can be parameterized but not the version we use on the site. No need to pretzel ourselves over this.

Sushisource (approved):

PR #415, opened 2025-03-02
pipe workflow logs to open telemetry

safareli:


interceptors-opentelemetry/src/worker.ts
```diff
@@ -17,7 +18,26 @@ function initializeRuntime() {
     // IMPORTANT: Make sure to configure the `telemetryOptions.logging` property
     //            below to also collect logs emitted by the native runtime.
     //
-    logger: new DefaultLogger('WARN'),
+    logger: new DefaultLogger("WARN", (entry) => {
+      const seconds = Number(entry.timestampNanos / BigInt(1e9));
+      const nanos = Number(entry.timestampNanos % BigInt(1e9));
+      const hrTime= [seconds, nanos];
```
stefanmatheis-labsruhr, line 24:
```suggestion
      const hrTime: HrTime = [seconds, nanos];
```

`hrTime` needed a proper type for typescript to be happy. 

```ts
import { HrTime } from "@opentelemetry/api"
```

PR #416, opened 2025-03-19
Background heartbeating sample

mnichols:
This shows (and tests) how to have a `heartbeat` executing inside a loop while performing a long running operation inside an Activity. 
This needs to support `cancellation` as well. 

## What was changed
Extended the `heartbeat` demonstration with a realistic long-running activity impl that needs to kick of both a periodic heartbeat and a long-running, blocking process inside an Activity.
For example, writing to a file might take a loooong time but I want to ping the Temporal service in the background while my sync file operation is doing it's thing.

## Why?
It should make clearer how to get these concurrent things working together.

## Checklist

2. How was this tested: `mocha` unit tests. Also verified cancellation works like I'd expect.

3. Any docs updates needed? Not really

### DIFF
--- a/activities-cancellation-heartbeating/package.json
+++ b/activities-cancellation-heartbeating/package.json
@@ -28,13 +28,16 @@
     "@temporalio/workflow": "^1.11.6"
   },
   "devDependencies": {
+    "@temporalio/testing": "^1.11.7",
     "@tsconfig/node18": "^18.2.4",
+    "@types/mocha": "^10.0.10",
     "@types/node": "^22.9.1",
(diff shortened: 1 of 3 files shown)


mnichols:
@mjameswh is there any feedback on this or should I close this and put this in my own repo showing common patterns?
The python samples has an `auto-heartbeater` sample so I thought this might be a compliment to that.

mjameswh:
> ###
> [May 13, 2025](#issuecomment-2877129102)

Sorry Mike, I have been very busy on other things. Please leave it open here, it is definitely a good fit for this repo. I will try to review sooner than later.

PR #423, opened 2025-06-24
Add testing helper to safe message handlers sample

yuandrew:
## What was changed
Added `maxHistoryLength` to safe message handlers sample 

## Why?
All of the other SDK samples have this and new docs point to it, we want this for consistency.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/message-passing/safe-message-handlers/src/cluster-manager.ts
+++ b/message-passing/safe-message-handlers/src/cluster-manager.ts
@@ -155,7 +155,7 @@ export class ClusterManager {
     }
     
     // This is just for ease-of-testing. In production, we trust temporal to tell us when to continue-as-new.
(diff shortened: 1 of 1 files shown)


message-passing/safe-message-handlers/src/cluster-manager.ts
```diff
@@ -25,15 +26,17 @@ export class ClusterManager {
   state: ClusterManagerState;
   seenJobs: Set<string>;
   nodesMutex: Mutex;
+  private maxHistoryLength: number | null;
```
dandavison, line 29: I think `maxHistoryLength?` would be appropriate rather than `| null`?

yuandrew: Thanks for the suggestion, clearly someone who's not the most familiar with TS writing TS (I also blame cursor 😊)

PR #452, opened 2026-03-09
Recreate sliding window sample in typescript. Based on Go/Java/Python samples

billrich2001:
## What was changed
 Made a new sample based on existing samples in other language sample repos

## Why?
Cloud customer asked if we had a typescript sample in a ticket


batch-sliding-window/src/workflows.ts
```diff
@@ -0,0 +1,224 @@
...
+  setHandler,
+  sleep,
+  startChild,
+  workflowInfo,
+} from '@temporalio/workflow';
+import type * as activities from './activities';
+
+const { getRecordCount, getRecords } = proxyActivities<ReturnType<(typeof activities)['createActivities']>>({
+  startToCloseTimeout: '5 seconds',
+});
```
chris-olszewski: A little cleaner IMO to avoid indexing into the module type
```suggestion
import type { createActivites } from './activities';

const { getRecordCount, getRecords } = proxyActivities<ReturnType<typeof createActivities>>({
  startToCloseTimeout: '5 seconds',
});
```

batch-sliding-window/src/workflows.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  const childPromises: Promise<number>[] = [];
+  let offset = 0;
+
+  for (let i = 0; i < input.partitions; i++) {
+    const childId = `${workflowInfo().workflowId}/${i}`;
+    let maximumPartitionOffset = offset + partitionSizes[i];
+    if (maximumPartitionOffset > totalRecords) {
+      maximumPartitionOffset = totalRecords;
+    }
```
chris-olszewski:
```suggestion
    const maximumPartitionOffset = Math.min(offset + partitionSizes[i], totalRecords);
```

chris-olszewski (approved):
Looks good to me. Just 2 small style comments.

PR #458, opened 2026-04-27
Nexus samples: use business IDs for workflow IDs

VegetarianOrc:
## What was changed

Stop using the Nexus request ID as the backing workflow ID. Build a meaningful business ID from the operation input.

## Why?

Our samples should reflect best practices around creating meaningful IDs. Part of https://github.com/temporalio/features/issues/692.

### DIFF
--- a/nexus-cancellation/src/service/handler.ts
+++ b/nexus-cancellation/src/service/handler.ts
@@ -6,7 +6,7 @@ import { helloWorkflow } from './workflows';
 /**
  * Creates a business-meaningful ID that is used to dedupe workflow starts
  * from the `helloService.hello` Nexus Operation.
- * 
+ *
  * @param input HelloInput
  * @returns A workflow ID derived from the Nexus Operation input.
  */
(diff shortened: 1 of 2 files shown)


chrsmith (approved):
I really like you introducing a standalone function `function workflowIdForHello(input: HelloInput): string {`. However, it seems like it would benefit from a doc comment... but that having been said, I have no idea what would be a reasonably clear description.

"Generates a meaningful business ID"
"Generates the Workflow ID, a semi-unique identifier for the workflow's invocation"?
🤷 😭

VegetarianOrc:
> I really like you introducing a standalone function `function workflowIdForHello(input: HelloInput): string {`. However, it seems like it would benefit from a doc comment... but that having been said, I have no idea what would be a reasonably clear description.
> 
> "Generates a meaningful business ID" "Generates the Workflow ID, a semi-unique identifier for the workflow's invocation"? 🤷 😭

Good suggestion. I added a doc comment that should help highlight the purpose!

THardy98 (approved):

PR #459, opened 2026-04-28
Fix husky pre-push

VegetarianOrc:
## What was changed

- Add functionality to support one level nesting organization of samples via HAS_CHILD_SAMPLES constant in .scripts/copy-shared-files.mjs.
- Run the pre-push hook that copies shared files el
- Add mocharc with large timeout to help reduce CI flakes.
- Update AI SDK tests with a deep equal and a skip if no OPENAI_API_KEY is set.

## Why?

The pre-commit was was failing locally due the `copy-shared-files.mjs` script failing because `./message-passing/package.json` does not exist. Fixing this bug uncovered a few lingering issues.

## Checklist

1. How was this tested:

- existing tests
- running the `copy-shared-files.mjs` script
- running the pre-push hook when pushing.


ai-sdk/src/activities.ts
```diff
@@ -2,7 +2,6 @@
 export async function getWeather(input: {
   location: string;
 }): Promise<{ city: string; temperatureRange: string; conditions: string }> {
-  console.log('Activity execution');
```
chris-olszewski, line 5: Reason for this removal?

chris-olszewski (approved):

ai-sdk/src/activities.ts
VegetarianOrc: Just was polluting the test output and decided to remove it.

PR #462, opened 2026-05-01
Added Standalone Activities sample

maciejdudko:
Added sample demonstrating use of Standalone Activities APIs.


standalone-activity/src/list.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  const { count } = await client.activity.count(query);
+  console.log(`Total activities: ${count}`);
+
+  console.log('ACTIVITY ID | RUN ID | ACTIVITY TYPE | STATUS | COMPLETED TIME');
+
+  for await (const a of client.activity.list(query)) {
+    console.log(
+      `${a.activityId} | ${a.activityRunId} | ${a.activityType} | ${a.status} | ${a.closeTime?.toISOString()}`,
+    );
```
THardy98, line 20: Nit: is there a reason we don't do:
```
const listActivities = await client.activity.list(query);
console.log(`Total activities: ${listActivities.length});

console.log('ACTIVITY ID | RUN ID | ACTIVITY TYPE | STATUS | COMPLETED TIME');

for (const a of listActivities) { ... }  
```

Could be `TOCTOU` between `client.activity.count` and `client.activity.list`

THardy98 (approved):

standalone-activity/src/list.ts
maciejdudko: We are demonstrating both the `list` and the `count` APIs.

PR #475, opened 2026-06-09
Fix pre-push hook when no terminal is attached

brianstrauch:
## What was changed

Pushing from non-interactive environments (Claude Code, IDE integrations, CI) failed in the pre-push hook for two reasons:

1. **`.husky/pre-push`**: `exec < /dev/tty` errors out when no TTY exists ("/dev/tty: Device not configured"). The redirect is now skipped when `/dev/tty` isn't readable.
2. **`.scripts/copy-shared-files.mjs`**: git feeds the pre-push hook its ref list on stdin, so without the redirect, the confirmation prompt read a ref line (e.g. `refs/heads/main ...`) as the answer — which isn't "Y" — and aborted the push. The script now only prompts when stdin is a terminal and proceeds with the default otherwise. Interactive pushes still get the prompt.

Also moves the pnpm-specific workspace settings from the root `.npmrc` into `pnpm-workspace.yaml` (pnpm 10's settings file, which npm never reads). npm doesn't understand those settings, so every `npm run` — including the hook — printed four "Unknown project config" warnings. Verified `pnpm install --lockfile-only` leaves `pnpm-lock.yaml` untouched, so resolution behavior is unchanged.

## Testing

Ran the hook exactly as git invokes it (ref data piped to stdin) through both the early-exit path and the full copy path: exit 0, no prompt, no warnings, no staged residue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Sushisource (approved):
Thanks this always drove me nuts

PR #483, opened 2026-06-24
AI-304: Layout-agnostic shared eslintrc; fix outdated post-create messages

xumaple:
Broaden the shared eslintrc workflow-override glob to `src/**/workflows*.ts` so the no-node-builtins rule covers nested per-scenario layouts (e.g. openai-agents' `src/<feature>/workflows.ts`), letting openai-agents use the shared eslintrc and drop out of `ESLINTRC_EXCLUDE`. All samples' eslintrc copies are re-synced.

Also fixes 8 samples (env-config, dsl-interpreter, eager-workflow-start, standalone-activity, state, nexus-cancellation, nexus-hello, nexus-messaging) whose `.post-create` told users to run npm scripts that don't exist (`start.watch`/`workflow`); each now matches its real scripts and is excluded from the shared post-create.

### DIFF
--- a/.scripts/copy-shared-files.mjs
+++ b/.scripts/copy-shared-files.mjs
@@ -42,7 +42,6 @@ const GITIGNORE_EXCLUDE = [
   'nestjs-exchange-rates',
 ];
 const ESLINTRC_EXCLUDE = [
-  'openai-agents',
   'nextjs-ecommerce-oneclick',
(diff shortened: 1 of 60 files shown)


brianstrauch (approved):
Interesting that this was a widespread issue. Thanks for fixing!

chris-olszewski (approved):

PR #488, opened 2026-06-29
AI-310: add streaming sample for @temporalio/openai-agents

xumaple:
Adds a streaming scenario to the openai-agents samples, demonstrating `run(agent, input, { stream: true })` over a Workflow Stream: a Workflow that hosts a `WorkflowStream`, a Worker configured with the streaming topic, and a client that subscribes to model stream events live from outside the Workflow.

Draft until `@temporalio/openai-agents` (with streaming) and `@temporalio/workflow-streams` publish — CI is red against the published packages until then.


openai-agents/src/streaming/worker.ts
```diff
@@ -0,0 +1,43 @@
...
+  const connection = await NativeConnection.connect({ address: 'localhost:7233' });
+  try {
+    const worker = await Worker.create({
+      connection,
+      taskQueue: 'openai-agents-streaming',
+      workflowsPath: require.resolve('./workflows'),
+      plugins: [
+        new OpenAIAgentsPlugin({
+          modelProvider: new OpenAIProvider({ apiKey }),
+          modelParams: { streamingTopic, streamingBatchInterval: '200 milliseconds' },
```
DABH: The worker is setting `streamingBatchInterval` here but the client just passes `{ streamingTopic }` - I think that means the interval doesn't actually get set (worker-side plugin `modelParams` don't enter the sandbox?).  Can you double check the examples for this?

openai-agents/src/streaming/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Run
+
+Run these from the `openai-agents/` root (run `npm install` there once first).
+
+```bash
+# In one terminal, start the Worker (requires a local Temporal server and OPENAI_API_KEY):
+OPENAI_API_KEY=sk-... npx ts-node src/streaming/worker.ts
+
+# In another terminal, start the streaming client:
+npx ts-node src/streaming/client.ts
```
DABH, line 29: The client needs OPENAI_API_KEY too (client.ts: 14-17), prefix with `OPENAI_API_KEY=sk-... `?

openai-agents/src/streaming/worker.ts
xumaple: config moved to the client plugin where it actually takes effect

openai-agents/src/streaming/README.md
xumaple: turned out the client never calls the model, so removed the key requirement rather than documenting it.

DABH (approved):
LGTM!

chris-olszewski (approved):

PR #495, opened 2026-07-13
Improve SANO sample

Quinn-With-Two-Ns:
## What was changed
Improve SANO sample

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):
I would make the default (fallback) namespace `my-caller-namespace` or `my-handler-namespace` instead of `default` but otherwise LGTM.

PR #506, opened 2026-08-06
fix(ai-sdk): upgrade to zod v4 and pin ai to 7.0.41

brianstrauch:
## Problem

The `ai-sdk` sample fails on a fresh install. Its `.npmrc` sets `package-lock=false`, so `npm install` floats every dependency to the newest in-range version. That surfaces two independent regressions.

### Bug A — TS2589 at compile time

`typescript` floats `^5.6.3` → 5.9.3, and TS 5.9 exceeds the instantiation depth limit on zod 3's types via the AI SDK v7 `tool()` helper:

```
src/workflows.ts(31,19): error TS2589: Type instantiation is excessively deep and possibly infinite.
```

Some users see an out-of-memory crash in `tsc` instead — that's the same problem, just dying before it can report the error. It is **not** caused by the `ai` version:

| typescript | zod | result |
|---|---|---|
| 5.9.3 | 3.25.76 | ❌ TS2589 |
| 5.7.3 (old lockfile) | 3.25.76 | ✅ |
| 5.9.3 | 4.4.3 | ✅ |

### Bug B — workflow tasks fail at runtime

Hidden behind Bug A; fixing zod alone exposes it. `ai` floats 7.0.36 → 7.0.55, pulling `@ai-sdk/provider-utils` >= 5.0.15, which added an unguarded module-scope call:

```js
var initialGlobalFetch = globalThis.fetch;
var initialGlobalFetchIsNodeDefault = isNodeDefaultFetch(initialGlobalFetch);

function isNodeDefaultFetch(fetch) {
  const source = Function.prototype.toString.call(fetch); // throws when fetch is undefined
```

Workflows are deterministic and have no `fetch`, so importing `ai` inside the sandbox throws at import time and every workflow task fails:

```
TypeError: Function.prototype.toString requires that 'this' be a Function
    at isNodeDefaultFetch (safe-node-fetch.ts:135:45)
    at importWorkflows (workflows-autogenerated-entrypoint.cjs:13:9)
```

Bisected to exactly `@ai-sdk/provider-utils@5.0.15`, first pulled in by `ai@7.0.42`.

## Fix

- `zod` → `^4.0.0` (resolves Bug A)
- `ai` → pinned `7.0.41`, the last release depending on provider-utils 5.0.14 (resolves Bug B)

`@ai-sdk/mcp` still resolves a 5.0.23 copy into the tree, but it is never evaluated in the sandbox, so it does not need pinning.

## Testing

All five samples (`haiku`, `tools`, `mcp`, `middleware`, `stream`) pass against a local dev server, via both pnpm and a fresh `npm install`, with zero worker errors. Build verified clean on both TS 5.7.3 and TS 5.9.3.

## Follow-ups (not in this PR)

- The `ai` pin is a **stopgap**. Bug B needs an upstream guard (`typeof fetch !== 'function'`) in `@ai-sdk/provider-utils`; an issue for `vercel/ai` is drafted but not yet filed.
- `package-lock=false` in `ai-sdk/.npmrc` is the root enabler — npm users get none of the reproducibility the committed `pnpm-lock.yaml` provides, so the sample can break again at any time from an unrelated upstream release.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ai-sdk/package.json
+++ b/ai-sdk/package.json
@@ -34,9 +34,9 @@
     "@temporalio/worker": "^1.22.0",
     "@temporalio/workflow": "^1.22.0",
     "@temporalio/workflow-streams": "^1.22.0",
-    "ai": "^7.0.0",
+    "ai": "7.0.41",
     "nanoid": "3.x",
-    "zod": "^3.25.76"
+    "zod": "^4.0.0"
   },
   "devDependencies": {
     "@temporalio/testing": "^1.22.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1197,7 +1197,7 @@ importers:
         version: 0.1.13(prettier@3.4.2)
       vercel:
         specifier: ^29.0.3
-        version: 29.4.0(@swc/core@1.10.11(@swc/helpers@0.5.15))(@types/node@22.12.0)(babel-plugin-macros@3.1.0)(terser@5.37.0)(ts-node@10.9.2(@swc/core@1.10.11(@swc/helpers@0.5.15))(@types/node@22.12.0)(typescript@4.9.5))
+        version: 29.4.0(@swc/core@1.10.11(@swc/helpers@0.5.15))(@types/node@22.12.0)(babel-plugin-macros@3.1.0)(terser@5.37.0)(ts-node@10.9.2(@swc/core@1.10.11(@swc/helpers@0.5.15))(@types/node@22.12.0)(typescript@5.7.3))
 
   food-delivery/apps/driver:
     dependencies:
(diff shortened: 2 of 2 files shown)


patbeqo (approved):

chris-olszewski (approved):

ai-sdk/package.json
```diff
@@ -34,9 +34,9 @@
     "@temporalio/worker": "^1.22.0",
     "@temporalio/workflow": "^1.22.0",
     "@temporalio/workflow-streams": "^1.22.0",
-    "ai": "^7.0.0",
+    "ai": "7.0.41",
```
moedash, line 37: I checked this on a fresh `npm install`. `ai@7.0.41` pins `@ai-sdk/provider-utils` at `5.0.14`, and npm nests that copy under `ai/`, so a floating sibling can't pull the sandbox up to a probe build. `@ai-sdk/mcp` does float in and drag `provider-utils@5.0.23` (the probe version) onto disk, but it stays worker-side and never reaches the workflow bundle, so that's fine I guess.

Two things still worth a thought:

Since the break hits every `@temporalio/ai-sdk` user who imports `ai` in a workflow, is a fix in `@temporalio/ai-sdk` (say a benign `fetch` in the sandbox) on the table, so samples don't each carry a pin? Is there an upstream issue we can link, so we know when to drop it?

`package.json` can't have comments, so the next person who wants to bump `ai` won't know why it's fixed. A one-line note in the README maybe?

ai-sdk/package.json
```diff
@@ -34,9 +34,9 @@
...
     "nanoid": "3.x",
-    "zod": "^3.25.76"
+    "zod": "^4.0.0"
```
moedash, line 39: Small one. The AI SDK packages peer on `zod: ^3.25.76 || ^4.1.8`. `^4.0.0` also lets `4.0.x` through `4.1.7`. Fresh installs float to newest, so they'd use `4.4.3` today and works. Do you want to set the floor to `^4.1.8`, so the declared range can't go below what the SDK supports?

pnpm-lock.yaml
```diff
@@ -32647,6 +32774,38 @@ snapshots:
 
   vary@1.1.2: {}
 
+  vercel@29.4.0(@swc/core@1.10.11(@swc/helpers@0.5.15))(@types/node@22.12.0)(babel-plugin-macros@3.1.0)(terser@5.37.0)(ts-node@10.9.2(@swc/core@1.10.11(@swc/helpers@0.5.15))(@types/node@22.12.0)(typescript@4.9.5)):
```
moedash: A chunk of this file isn't about the zod bump. `vercel`, `@trpc/next`, `@vercel/remix-*`, and `ts-node` re-resolve onto `typescript@4.9.5` in the `food-delivery` workspace, which is unrelated to `ai-sdk`.

Looks like a local pnpm version that differs from CI's (`pnpm/action-setup` pins v10). Could you regenerate the lockfile with pnpm 10, so the diff stays scoped to the `ai-sdk` importer? CI runs plain `pnpm install`, not `--frozen-lockfile`, so it's low risk. But the extra churn is hard to review, and it quietly shifts what the other samples resolve.

ai-sdk/package.json
brianstrauch: I forget the reason for why we went with ^4.1.8 for the SDK, but you wouldn't actually resolve to something between 4.0.0 and 4.1.8 because of the constraint on the SDK, which you're always using for these samples, so I don't feel the need to keep samples dependencies in lock-step with the SDK

pnpm-lock.yaml
brianstrauch: Good find!

ai-sdk/package.json
brianstrauch: https://github.com/vercel/ai/issues/18528

I'll make a note to bump the version once this is fixed! Hopefully we can automate this in the future.

PR #512, opened 2026-08-24
AI-283: Add Google ADK samples

xumaple:
Eight samples demonstrating how to run Google ADK (`@google/adk`) agents as durable Temporal Workflows with `@temporalio/google-adk-agents`: multi-turn agent chat, deterministic and Activity-backed tools, multi-agent orchestration, stateless MCP, token streaming, human-in-the-loop, structured output, and observability. Each model-dependent scenario has an API-key-free local mode, and all scenarios have Worker tests that run without an API key.

Note: `@temporalio/google-adk-agents` is not published yet, so CI installation will fail until it ships.


brianstrauch (approved):

DABH (approved):
LG once CI passes

chris-olszewski (approved):

PR #513, opened 2026-08-24
Adding nexus TemporalOperationHandler to nexus-messaging examples

JoshuaFrenchwood:
## What was changed

Updating examples to use TemporalOperationHandler for nexus messaging.

## Checklist

1. Closes 

2. How was this tested:

Ran the examples and verified with expected output

3. Any docs updates needed?

### DIFF
--- a/nexus-messaging/src/callerpattern/README.md
+++ b/nexus-messaging/src/callerpattern/README.md
@@ -22,6 +22,7 @@ Start a compatible Temporal dev server with Workflow Update callbacks enabled:
 ./temporal server start-dev \
   --dynamic-config-value history.enableCHASMCallbacks=true \
   --dynamic-config-value history.enableUpdateCallbacks=true \
+  --dynamic-config-value history.enableCHASMSignalBacklinks=true \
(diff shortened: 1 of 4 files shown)


Quinn-With-Two-Ns:
I think we need some DC configs too to make signal links work at least , I need to double check which exactly

nexus-messaging/src/callerpattern/README.md
```diff
@@ -16,19 +16,23 @@ The caller workflow:
 
 ### Running
 
-Start a Temporal server:
+Start a compatible Temporal dev server with Workflow Update callbacks enabled:
 
 ```bash
-temporal server start-dev
+./temporal server start-dev \
```
Quinn-With-Two-Ns, line 22: I think we would also need `history.enableCHASMSignalBacklinks`

JoshuaFrenchwood: Thanks, updated

Quinn-With-Two-Ns (approved):

PR #518, opened 2026-09-09
Remove obsolete content

tomwheeler:
## What was changed

I removed the section at the top. I also removed the unnecessary periods at the end of the bullet points.

## Why?

The VSCode plugin link and blog post both lead to a 404 page. The VSCode extension is no longer maintained.

I removed the periods at the end of the bullet points because they are unnecessary. The bullet points are not complete sentences, so they should not have periods.

### DIFF
--- a/vscode-debugger/README.md
+++ b/vscode-debugger/README.md
@@ -1,12 +1,6 @@
-# VS Code Debugger
-
-Use our [VS Code plugin](https://marketplace.visualstudio.com/items?itemName=temporal-technologies.temporalio) to debug a running or completed Workflow Execution. See this blog post for a walkthrough:
-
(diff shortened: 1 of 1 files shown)


mjameswh:
For the record, the PR as it currently stands is incomplete. It keeps the `vscode-debugger` sample, but its `README.md` file no longer has a proper title, and we're left with incomplete instructions on how to use the sample (how do you install the vscode extension?)

I think we really have two avenues going forward. Either:
1. We keep the sample, fix the broken URL, and put some real efforts in maintaining the debugger extension.
2. We completely remove the sample.

Not something I can decide alone. In fact, I even think that may no longer fall under the SDK team's umbrella. Let's discuss internally.

mjameswh:
Confirmed internally. Superseded by #524 for complete and proper removal of the sample.

PR #523, opened 2026-09-16
Restore SDK ownership of Nexus samples

mjameswh:
## What changed

PR #517 added a Nexus-only CODEOWNERS rule for `/nexus*/`.

Because later matches replace the default owner set, that rule prevents an SDK-team approval from satisfying repository-wide maintenance changes such as PR #522.

Make the SDK and Nexus teams co-owners of Nexus sample directories. This preserves Nexus review coverage while keeping the SDK team able to maintain the complete samples repository, consistent with the existing AI sample rules.

Validation:

- `git diff --check passes`
- every active CODEOWNERS rule includes `@temporalio/sdk`


mjameswh (commented):
Added an explicit warning to avoid future reoccurence.

tconley1428 (approved):

